PR #5, opened 2016-11-22
need the extra # for title to render correctly

mairaw:
it would probably be good to search for this issue across the migrated content


mairaw:
ah never mind. just noticed that @svick already caught that with PR #3

PR #6, opened 2016-11-22
Add Changelog for Visual Studio Tools for Unity (VSTU) v2.8 (3.0 Preview)

sailro:
Added changelog entry for the latest 2.8 release.

Added "real" versions for other previews. This version number is used on the gallery, for assembly versioning and within the Unity about window.

cc @jbevain


gregvanl:
We're currently blocked from publishing due to an infrastructure issue.  I expect this won't be published live until after the Thanksgiving holidays.

PR #8, opened 2016-11-24
Correct the description of the command prompt

SteveDunn:


docs/install/create-an-offline-installation-of-visual-studio.md
```diff
@@ -40,12 +40,12 @@ You can create an offline installation layout that contains all the files you ne
 
 #### To create an offline installation layout of Visual Studio
 1. Download the Visual Studio setup executable file to a drive on your local machine.
-  For example, [download the vs_enterprise.exe file](https://www.visualstudio.com/vs/visual-studio-2017-rc/).
-2. Run `vs_enterprise.exe` with the following arguments (switches) from a command prompt:
+  For example, [download the vs_enterprise.exe file](https://www.visualstudio.com/vs/visual-studio-2017-rc/).  It doesn't have to be the Enterprise edition, it can be the Community or Professional editions.
+2. Run `vs_enterprise.exe` with the following arguments (switches) from a command prompt **with Administrator rights**:
 
-   a. Add `--layout <path>`, where `<path>` is the location where you want the layout to download to. By default, all languages are downloaded. (See Example A.)
+   a. Add `--layout <path>`, where `<path>` is the location where you saved the exe in step 1.
```
timsneath: Thanks for your contribution. But this piece isn't quite right, @SteveDunn: the layout path represents the location where the download will be cached. It doesn't have to be the location where you downloaded the initial setup executable.

TerryGLee: Re: the additional text about the editions...we appreciate your feedback, @SteveDunn. We'll take that under advisement and report back with next steps.

TerryGLee: Also, @SteveDunn, we appreciate your input on the example text about all languages being downloaded. We will address that slightly differently than how you've suggested; I'll report back after the work is done. (At that point, we'll most likely close this merge request so that it doesn't conflict with other text changes that are in the queue.)

TerryGLee: @SteveDunn - We're going to opt not to mention editions at this time, and we don't think it's a good idea to mention again that all languages are downloaded again in Step 2b, as that applies directly to Example A and therefore is mentioned in Step 2a. We thank you for submitting this PR, but we're closing it for now.

PR #10, opened 2016-11-29
Add info to change installer UI language

artl93:
The installer itself can have its UI language changed after the fact.

### DIFF
--- a/docs/install/install-visual-studio.md
+++ b/docs/install/install-visual-studio.md
@@ -94,10 +94,9 @@ To install Visual Studio 2017 RC in a language of your choosing, click the **Lan
   > [!IMPORTANT]
   > While Visual Studio 2017 RC in general is supported for use in a production environment, those workloads and components that are marked "Preview" in the installation UI are not supported for use in a production environment.
   
-## Changing the installer Language
+## Changing the installer language
 
-By default, the installer program tries to match the language of the operating system when run for the first time. The installer will remember this setting. The setting can later be chagned by running the installer from the command line. For example, the user can force the installer to run in Englishby running **vs_installer.exe --locale en-US"**. This setting will then be remembered when the installer is run the next time. 
-The installer supports the following language tokens: zh-CN, zh-TW, cs-CZ, en-US, fr-FR, de-DE, it-IT, ja-JP, ko-KR, pl-PL, pt-BR, ru-RU, es-ES, and tr-TR. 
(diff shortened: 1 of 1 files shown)


docs/install/install-visual-studio.md
```diff
@@ -93,6 +93,11 @@ To install Visual Studio 2017 RC in a language of your choosing, click the **Lan
 
   > [!IMPORTANT]
   > While Visual Studio 2017 RC in general is supported for use in a production environment, those workloads and components that are marked "Preview" in the installation UI are not supported for use in a production environment.
+  
+## Changing the installer Language
```
gregvanl: No need to capitalize "Language" here.  Can just be "Changing the installer language"

artl93:
I was still working on it... :-(

mairaw:
As a suggestion, for .NET we have a label called WIP to indicate topics are not ready to be merged yet.

docs/install/install-visual-studio.md
TerryGLee: This should also be a level 3 heading, as it is a child of the parent, "Install language packs"
Also, please change this to "Change the installer language," because it's a best practice to avoid gerunds (words ending in '-ing' whenever we can.

docs/install/install-visual-studio.md
```diff
@@ -93,6 +93,11 @@ To install Visual Studio 2017 RC in a language of your choosing, click the **Lan
...
+
+By default, the installer program tries to match the language of the operating system when run for the first time. The installer will remember this setting. The setting can later be chagned by running the installer from the command line. For example, the user can force the installer to run in Englishby running **vs_installer.exe --locale en-US"**. This setting will then be remembered when the installer is run the next time.
```
TerryGLee: As well, I'd like to word-craft this a bit:

By default, the installer program tries to match the language of the operating system when it is run for the first time. The installer will remember this setting. You can change the setting later by running the installer from the command line. For example, you can force the installer to run in English by running the **vs_installer.exe --locale en-US"** command. This setting will then be remembered when the installer is run the next time. 

It's best to say "you" when you can, as a best practice.

artl93:
Not sure why your commenting on an closed pull request with unrelated changes.


  1.  Each pull request and commit should be always ONE complete change.
  2.  When the pull request is closed, the comments left are absolutely ignored because they cannot be acted upon. The request is closed.

artl93:
I still think this guidance is horribly unprofessional for technical documentation. This guidance must be used in an indeterminate sense: You can.. you should… you may… Otherwise it supposes too much of the reader and otherwise sounds like something I was never allowed to submit in any paper past the seventh grade.

PR #17, opened 2016-12-08
bad link in line 73

kynowoo:
It may means "```[DocFX](https://dotnet.github.io/docfx/)```" for "```[DocFX](http://dotnet.github.io/docfx/)```"


gregvanl:
Thanks!

PR #24, opened 2017-01-23
fixing metadata section

mairaw:
this topic showed up in the .NET duplicate title report because of issues with the metadata values


mairaw:
moving this to the private repo

PR #27, opened 2017-01-27
Update use-command-line-parameters-to-install-visual-studio.md

Huy-Nguyen-MSFT:
Fix the documentation for --add and --remove to make it clear that [--add Workload1 Workload2] is not valid. The valid argument is [--add Workload1 --add Workload2].


TerryGLee:
Huy, thanks for your suggestion. We'll take your suggestion under consideration. And, we'll report back to this thread soon.

gregvanl:
Hi @Huy-Nguyen-MSFT, Elliot has added quite a bit of content to the --add and --remove table entries so I think it would be better to move your changes to a Note below. See https://github.com/Microsoft/vsdocs/commit/3d668d232013bda43669c685fabde3ff0b832a31 and let me know if you have any questions or comments.
Include @TerryGLee for visibility.

PR #38, opened 2017-02-15
Improve C++ string format specifier documentation

gregg-miskelly:
's8' documentation didn't tell users what it actually did
's*b' documentation was inconsistent
No documentation for ',env'

### DIFF
--- a/docs/debugger/format-specifiers-in-cpp.md
+++ b/docs/debugger/format-specifiers-in-cpp.md
@@ -85,12 +85,13 @@ int main() {
 |X<br /><br /> **H**|hexadecimal integer|102|0xCCCCCCCC|  
 |c|single character|0x0065, c|101 'e'|  
 |s|const char* string|\<location> “hello world”|"hello world"|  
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@rchiodo FYI. Also, should we remove the 'Specifiers in bold are not supported for interop debugging with C++/CLI.' part of the documentation now?

gregg-miskelly:
Here is a link to the updated documentation in formatted form: https://github.com/gregg-miskelly/visualstudio-docs/blob/master/docs/debugger/format-specifiers-in-cpp.md#-format-specifiers

rchiodo:
Sorry not sure. I'm assuming you asking me this because the same format specifiers should be supported in C++/CLI as native is, but I can't answer that question. At least not without a bunch of research. I suppose it would just be quicker to try it and see what happens.

gregg-miskelly:
@rchiodo Correct. I am assuming that comment is now wrong, because it was only true because we were using mcee.dll, and now that we are using CppEE for native data it will not be true. But I don't know for sure.

PR #50, opened 2017-03-13
Fix component ID

drewnoakes:
This document uses two different component IDs for _Visual Studio core editor_.

From my testing, it seems that `Microsoft.VisualStudio.Component.CoreEditor` is the correct one, not `Microsoft.VisualStudio.CoreEditor`.


gregvanl:
Thanks!

PR #61, opened 2017-03-28
Update with steps for administrative deployment

timsneath:
I've updated this section based on customer feedback to clarify the various steps required to deploy Visual Studio in an enterprise environment.


timsneath:
Sounds good. We'll fix this limitation in a forthcoming servicing update, at which point we can remove that caveat. Thanks for pushing live!

PR #68, opened 2017-04-01
Update getting-started-with-visual-studio-tools-for-unity.md

angerico-carino:
Just added an information box for users with Visual Studio 2017.


TerryGLee:
Thanks, @angerico-carino. We have plans to add further information about this very subject to this page, but we'll start with your Note. 

Thanks for contributing to our documentation!

~ @TerryGLee

PR #71, opened 2017-04-06
Bump Visual Studio Tools for Unity changelog to 3.1.0.0

sailro:


sailro:
cc @jbevain @TerryGLee

sailro:
We'll be able to publish the bits this morning, so we could change the date to 4/6

TerryGLee:
Let's publish it as is and if we want, we can change the date to 4/6/2017 after-the-fact.

sailro:
Bits are live

PR #75, opened 2017-04-10
Update create-an-offline-installation-of-visual-studio.md

oscarcao:
Some clarifications regarding the download process and filenames described in the guide vs the actual downloaded file filenames.


oscarcao:
Fixed typo.

TerryGLee:
We are in the midst of revising all the installation content. We will include edited versions of your updates in the revised set of installation content, which should be published in the next two weeks.

PR #86, opened 2017-04-18
Add TestPlatform log info

genlu:
@ManishJayaswal @rpetrusha


rpetrusha:
Actually, it seems that I can do it.

PR #99, opened 2017-04-26
Update download links with direct link to product

kelltrick:
This fixes an ambiguity in the documentation where links from https://www.visualstudio.com/downloads?utm_source=mscom&utm_campaign=msdocs would produce an exe file with a number appended to the end. These links download the bootstrapper as they appear further down in the document

### DIFF
--- a/docs/install/create-an-offline-installation-of-visual-studio.md
+++ b/docs/install/create-an-offline-installation-of-visual-studio.md
@@ -40,17 +40,17 @@ We understand that a lot of customers want an offline installer for [Visual Stud
 Here's how.
 
 ## Download the setup file you want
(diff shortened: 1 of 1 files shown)


kelltrick:
I've update the PR to include merging the link and file name columns as suggested by @TerryGLee . The table looks cleaner now without the full download link.

timsneath:
Even nicer!

kelltrick:
The reason to leave the ID off the end of the bootstrapper is that if users copy and paste commands directly from this document, the bootstrapper exe won't match. I believe that smoothing out any kinks between the documentation and the actual process involved in doing this out weigh the need for the ID.

PR #105, opened 2017-05-02
Master pull request for installer docs changes

timsneath:
This pull request contains a number of new pages and updates to the installer documentation.


timsneath:
Thanks Terry! I assume you’ll do a grammar and text pass when they get onto the production server? I’ll do another content pass at the same time.

PR #130, opened 2017-05-23
Updated Mono Uninstallation Command

junian:
Added `sudo rm -rf /etc/paths.d/mono-commands` to Mono uninstallation command. This is based on official Mono website: http://www.mono-project.com/docs/about-mono/supported-platforms/osx/


kempb:
@amburns

PR #150, opened 2017-06-12
Fix Directory.Build.targets name

jeffkl:
For this page, the file name should be `Directory.Build.targets` not `Directory.Build.props`


jeffkl:
Please review @Microsoft/msbuild-maintainers

PR #165, opened 2017-06-21
Changed "var" to "const"

EricWiener:


ghogen:
@MikeJo5000 can you take a look?

PR #309, opened 2017-10-06
deleted a duplicate link

gsamat:
there was a duplicate link in the page, leading to the page itself, I just deleted it


kraigb:
Thanks for catching this!

PR #316, opened 2017-10-11
Update Dotfuscator CE links

Joe-Sewell-PreEmptive:
Some restructuring on the Dotfuscator CE User Guide at preemptive.com changed some URLs. While there are currently redirects in place, this commit updates the links to use the new URLs.


Joe-Sewell-PreEmptive:
Thanks, @TerryGLee !

PR #329, opened 2017-10-22
Replaced period by Oxford comma

bartlannoeye:


yesman85:
I believe no comma should be added her at all.

bartlannoeye:
That's the whole discussion about Oxford comma :)
Either way, it shouldn't be a period.

PR #332, opened 2017-10-25
Fix Windows Server typo

Megasware128:
Change Windows Server 2017 to 2016


gewarren:
@Megasware128 thank you for fixing that!

PR #341, opened 2017-11-07
Typo check, s/effect/affect/

kkm000:
LSL may *affect* Visual Studio extensions, but cannot bring them about out of nowhere! :)


gregvanl:
Hi @kkm000, thank you for offering a correction. Lightweight solution load is going to be deprecated shortly and this entire topic will be removed next week.

PR #344, opened 2017-11-08
adding alternate reporting stub for those...

StingyJack:
... of us who get stuck and cant complete the installation.


TerryGLee:
Confirmed with suggested changes and follow-up tasks as described in this PR.

PR #339, opened 2017-11-02
Add known issue to debug-live-azure-applications.md

leftler:
As mentioned [in the comments of the article](https://docs.microsoft.com/en-us/visualstudio/debugger/debug-live-azure-applications) by Nikhil_Joglekar_MSFT on Oct 12, 2017, using snappoints is not currently supported for API App resources.

The documentation should be updated to show that information till a version is released that does work with Api Apps.

PR #350, opened 2017-11-15
Add lost known issue from transfer.

leftler:
The known issue bullet point `- Snapshot debugging API Apps is not yet supported - they will be supported in a future version of VS.` that was added in from #339 was lost during the transfer of the known issues section from `debug-live-azure-applications.md` to `debug-live-azure-apps-troubleshooting.md` during the 441cb8db6e6e425cddbc5344edf7855e027ab7f5 commit.


leftler:
@Mikejo5000 would you take a look at this since you are the one who approved #339 and did the 441cb8d commit.

PR #357, opened 2017-11-20
Fix broken link

LazarusX:


gregvanl:
@TerryGLee It looks like this PR may no longer be relevant as the content has changed?

TerryGLee:
I've verified that the bookmark link is indeed broken on the updated page. Thanks @LazarusX for noticing and creating an PR. We'll get this processed.

PR #363, opened 2017-11-30
Remove paragraphs regarding [un]registration

heku:
The paragraphs seems shouldn't be here, and they already exist on page https://docs.microsoft.com/en-us/visualstudio/extensibility/registering-and-unregistering-vspackages


gewarren:
@gregvanl can you review please?

gregvanl:
Thanks for spotting the duplicate content.

PR #373, opened 2017-12-05
Update of file path name for background image

samchivers:
The background image provided is called `WPF_watermark.png` but the code calling it was looking for a file called `watermark.png` - updated for consistency.


gewarren:
@samchivers thank you for doing this. I think the author of the topic intentionally named the image 'watermark.png'. Perhaps a better resolution is to change the alternate text for the image to "watermark.png" instead of "wpf_watermark.png". Would that solve the discrepancy as you see it?

gewarren:
Updated the title for the image for when you hover over it with a mouse.

samchivers:
@gewarren Sorry for the delayed response, fix looks great. All the best, Sam

PR #383, opened 2017-12-14
More realistic flow of steps, with project template detail

peteraritchie:


gewarren:
@peteraritchie thanks for doing this. i'll take a look at it.

PR #411, opened 2017-12-29
changed classes from internal to public

nhambayi:


gregvanl:
Hi @nhambayi, I'm not sure the classes in this example need to be changed from internal to public. Can you provide more details on why you want to make this change to the example.

gregvanl:
No reply back from PR creator after 10 days so closing.

PR #421, opened 2018-01-03
added one more OOP process name

heejaechang:
depends on arch (x86 or AnyCpu) of out of proc, name can be either ServiceHub.RoslynCodeAnalysisService32.exe or 
 ServiceHub.RoslynCodeAnalysisService.exe


heejaechang:
not sure how I should tag for review? @gregvanl ?

heejaechang:
by the way, I own the ServiceHub.RoslynCodeAnalysisService(32) process.

PR #432, opened 2018-01-08
Fix errors

r0bfr:
- The "How to" says that the Wizard window will close when the button is clicked.  However, the actual Close() call was missing from the button's click handler.

- The switches in the sn.exe command snippets had spaces after the hyphen character, which results in an error when the command is copied/pasted.


gregvanl:
Thanks!

PR #448, opened 2018-01-10
Update note on Accessibility Levels in editorconfig naming

abethcrane:
The note incorrectly said that an accessibility_level must be specified. This led me down the wrong path, and it wasn't until I removed the line that my parameter naming rule began to work. I've updated the note to indicate this.


gewarren:
@abethcrane Thank you for pointing this glitch out. I modified your suggested text a little and it should be live tomorrow.

PR #473, opened 2018-01-19
Lsp grammar/style

bplubell:
Parts of the "How work on the LSP started" section were unreadable. Theses changes should help convey what was originally intended. There is some inference on my part (based on context) in the paragraph that talk about JSON-RPC -- if someone could review it for accuracy, that would be appreciated.


gregvanl:
Thanks!

PR #485, opened 2018-01-24
Update finding-and-using-visual-studio-extensions.md

PooyaZv:
This is to add information related to the new Extension UI-Delay notifications (externally "Unresponsiveness Notifications"), which is similar to Extension Crash Notifications but shows notifications for extension causing VS UI to be unresponsive. Michael (author of Crash notifications and I) wanted to put these 2 under one node because of the shared functionality as well as similarity in nature.


gewarren:
@PooyaZv I'm a bit confused why this is here. Did you create a PR to each repo? You only need to create a PR in the private (-pr) repo and it will replicate automatically to this one. I'm closing this one for now.

PR #591, opened 2018-02-20
Mention name mangling in MSBuild solution builds

rainersigwald:
MSBuild replaces some characters in project names when creating targets

But this wasn't mentioned in the documentation.

Also mention troubleshooting options.
for individual projects in a solution metaproject:

https://github.com/Microsoft/msbuild/blob/adf225bf6abf2c6ab41a3214f6806b7618878939/src/Build/Construction/Solution/ProjectInSolution.cs#L397

Based on customer feedback sent via email.


rainersigwald:
@andygerlicher, @Mikejo5000 Does this look like good phrasing?

Mikejo5000:
@rainersigwald A made a couple of minor edits to the phrasing.

PR #608, opened 2018-02-24
Correct typo in C# ("aync" => "async")

GregTrevellick:
'asyncServiceProvider' is created, and used in the subsequent line of code, but the subsequent line of code is missing the first letter 's' i.e. it is 'ayncServiceProvider' not 'asyncServiceProvider'


gregvanl:
Thanks!

PR #631, opened 2018-03-01
Removes dashes from the sample code

AmadeusW:
Removes dashes from the sample code so that it can be easily copy-pasted.

(i'm not sure why it makes line 77 as edited, I tried to make this PR a few times and it keeps marking this line as changed)


gewarren:
@AmadeusW thanks for your contribution! It is normal for the last line to appear changed in the diff; I'm not sure why.

gregvanl:
Thanks!

PR #636, opened 2018-03-02
Fix typos in frequently used keyboard commands

flabby99:
Some commands were missing their leading character. For example dit.SurroundWith should be Edit.SurroundWith, which I have changed it to.


gewarren:
@flabby99 thanks so much for contributing and fixing this!

PR #694, opened 2018-03-14
changed "15.5" to "(latest version)"

PoulChapman:


gewarren:
@PoulChapman should we also remove the word "version" so it doesn't say Visual Studio 2017 version (latest version)?

PR #707, opened 2018-03-18
Update the sample to show correct thread transitions

BertanAygun:
Updated the sample for service initialization to showcase how service initialization work should be divided between background worker threads and Visual Studio main thread.


BertanAygun:
Abandoned without completing commit.

PR #715, opened 2018-03-20
C-Card Alt Tags & Accessibility: removed alt content

cmcclister:
Alt tags with values are breaking accessibility when used in the C-Cards.


cmcclister:
#sign-off apexprod

gewarren:
@cmcclister can you explain why you did this so we know for the future?

cmcclister:
@gewarren alt tags with values are breaking accessibility when used in the C-Cards, per @adkinn

PR #722, opened 2018-03-21
Older projects need a different base docker image : microsoft/dotnet-framework

bugoutfr:
Hi,

I added a note regarding an error which arises when building projects based on an older version of the Framework if you don't use a dotnet-framework base image (instead of windowservercore image) :

“C:\sources\Mysolution.sln” (default target) (1) ->
“C:\sources\Project.Web\Project.Web.csproj” (default target) (2) ->
“C:\sources\Project.DataAccess\Project.DataAccess.csproj” (default target) (3:2) ->
(CoreCompile target) ->
C:\BuildTools\MSBuild\15.0\bin\Roslyn\Microsoft.CSharp.Core.targets(84,5): error MSB6003: The specified task executab
le “csc.exe” could not be run. Could not load file or assembly ‘System.IO.FileSystem, Version=4.0.1.0, Culture=neutral,
PublicKeyToken=b03f5f7f11d50a3a’ or one of its dependencies. The system cannot find the file specified. [C:\sources\Project.DataAccess\Project.DataAccess.csproj]

### DIFF
--- a/docs/install/build-tools-container.md
+++ b/docs/install/build-tools-container.md
@@ -158,7 +158,7 @@ You must save the following example Dockerfile to a new file on your disk. If th
    ```
 
 > [!NOTE]
-> If you are using Visual Studio 2017 version 15.6 or later, you should use the following as Docker image as the base image instead: microsoft/dotnet-framework
+> If you are using Visual Studio 2017 version 15.6 or later, you should use the following Docker image as the base image instead: microsoft/dotnet-framework
 
 ```dockerfile
   # Use the following image to build older projects
--- a/docs/install/build-tools-container.md
+++ b/docs/install/build-tools-container.md
@@ -158,7 +158,7 @@ You must save the following example Dockerfile to a new file on your disk. If th
    ```
 
 > [!NOTE]
-> If you want to build solutions / projects referencing older versions of the .NET Framework (e.g. v4.0) you should use the following Docker image instead: microsoft/dotnet-framework
+> If you are using Visual Studio 2017 version 15.6 or later, you should use the following as Docker image as the base image instead: microsoft/dotnet-framework
 
 ```dockerfile
   # Use the following image to build older projects
--- a/docs/install/build-tools-container.md
+++ b/docs/install/build-tools-container.md
@@ -157,6 +157,14 @@ You must save the following example Dockerfile to a new file on your disk. If th
    CMD ["powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
    ```
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


TerryGLee:
@heaths Would you take a look at this PR, please?

bugoutfr:
Hi @heaths  

Thanks for your answer.

I am not sure I understand your point though. In my case I used this article in order to do what you just said: install the build Tools into a container. And as a result of it I ended up with a useless Docker Image  because it simply couldn't compile anything referencing .NET Framework 4 (or 4.5). I spent almost a week trying to find a solution and rebuilding numerous versions of this image until I found found out that referencing a microsoft/dotnet-framework image instead could solve the problem.

My point here is not to explain how to use the build tools or how to build .NET solutions / projects, but to avoid others the hassle of running into the same problem when a simple line of code could save their day. I posted comments about this on your blog two weeks ago, but they are not visible yet (awaiting moderation?)

The modifications I'm suggesting here are not related in any way with the .NET Code SDK you are talking about.

Regards,
Simon

heaths:
The purpose of the article is to inform how to install Build Tools on Windows. Using the [microsoft/dotnet-framework](https://hub.docker.com/r/microsoft/dotnet-framework/) images as a base does mean the .NET Framework is pre-installed, but looking at the tags appears less obvious what `:latest` is relative to the base microsoft/windowsservercore tag.

Build Tools will install the .NET Framework but, as I noted, does have a problem we are all investigating right now when trying to install on a container. I.e. this is a current bug that will be fixed. Your statement reads like it implies this is necessary. Instead, perhaps noting that as of Visual Studio 2017 version 15.6 this requires starting from such a base image (i.e. the note survives and applies to older images well past when this bug is fixed) it would make more sense.

bugoutfr:
That's much clearer now, I altered the file following your recommendation. Don't hesitate to rephrase it if necessary.
Thanks for the explanation.

heaths:
Since this will be fixed at some point, we'll have to remember to update the docs. I will discuss with the .NET Framework team and make an appropriate change.

bugoutfr:
Yes, it's temporary. Thanks.

heaths:
@TerryGLee, if you want to close this one out I'll have a change coming very soon with recommendations from the .NET Framework team that might include more changes than requested here.

TerryGLee:
@bugoutfr Thanks for your great feedback, suggestions, and dialog. We are going to close this PR now, as @heaths will be updating the content soon (per his earlier comment).

TBBle:
@heaths Was this issue ever fixed, and if so, in what releases?

PR #742, opened 2018-03-25
deleted unnecessary require('http')

user135711:
This is a great starter template, but the line "var http = require('http');" is unnecessary because express created its own server and the other server never gets used for possible scenarios where it would be necessary(https://stackoverflow.com/questions/17696801/express-js-app-listen-vs-server-listen). I propose removing it.


gewarren:
@user135711 thanks for contributing. @Mikejo5000 can you look at this?

Mikejo5000:
@user135711 Looks good. Thanks!

PR #743, opened 2018-03-25
MSBuild: Additional forbidden properties

rainersigwald:
These names are disallowed because they conflict with MSBuild language
elements, but they weren't listed here.

See the official list:
https://github.com/Microsoft/msbuild/blob/d83a9eda20970d2bfa3ad96d47cbecf93343256b/src/Shared/XMakeElements.cs#L41-L58

Closes Microsoft/msbuild#3133.


rainersigwald:
cc @Mikejo5000 

Thanks for pointing this out, @dazinator!

PR #759, opened 2018-03-29
Add FilePath search box FAQ

kendrahavens:
With VS 2017 Update 15.7 Preview 3 there will no longer be a File Path search option in the Test Explorer. It was taken out due to low usage and the performance improvement seen without it.


gewarren:
@kendrahavens do you want this to go live next Wednesday, or now?

PR #785, opened 2018-04-06
Fix name of referenced group

r0bfr:
By default the template creates a group named "MyMenuGroup".  Not one named "MyGroup".


Bethannfrum:
Fix and return to oig format

gregvanl:
Thanks!

PR #786, opened 2018-04-06
Add new help page for process exit during func eval

jsschultz:
In some rare cases, evaluating a func eval can result in the abrupt
termination of the debuggee process. When that happens, we will show an
error dialog that will provide a help button to find out more details on
how to prevent/diagnose this issue.

### DIFF
--- a/docs/debugger/error-evaluating-the-function-process-exited.md
+++ b/docs/debugger/error-evaluating-the-function-process-exited.md
@@ -1,0 +1,50 @@
+---
+title: "Error: The target process exited while evaluating the function &#39;function&#39; | Microsoft Docs"
+ms.custom: ""
+ms.date: "4/06/2018"
+ms.reviewer: ""
+ms.suite: ""
+ms.tgt_pltfrm: ""
+ms.topic: "reference"
+f1_keywords: 
+  - "vs.debug.error.process_exit_func_eval_abort"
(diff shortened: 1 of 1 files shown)


Mikejo5000:
@jsschultz Hi John, so the Help button in the dialog box is wired up to link to this page? What kind of link is it (I assume it's an FwLink that opens the page URL)? It's possible that the F1 keyword would be used if someone presses F1 in the error dialog box (in which case the F1 keyword would need to match the one you set here in the metadata), but that would need to be wired up in the IDE.

jsschultz:
@Mikejo5000 I'm trying to hunt down where the link is generated, but here is the url it opens the browser with: https://docs.microsoft.com/en-us/visualstudio/index?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(vs.debug.error.process_exit_during_func_eval)%3Bk(TargetFrameworkMoniker-.NETFramework%2CVersion%3Dv4.6)%3Bk(DevLang-csharp)%26rd%3Dtrue

Does that look right?

Mikejo5000:
@jsschultz Yes, that is an F1 link, so that ought to work fine. You may not need to hunt it down.

PR #791, opened 2018-04-10
Fix mistake in TEST_CLASS_INITIALIZE explanation

malchera:
"Defines methodName as a method that runs BEFORE each test class is created" - not "after"


gewarren:
@mikeblome can you look at this small change?

PR #816, opened 2018-04-16
MSBuild: document "magic" extensibility

rainersigwald:
Documentation for extensibility mechanisms available in MSBuild's default props/
targets that are non-obvious.

Closes Microsoft/msbuild#3166.

@Mikejo5000

### DIFF
--- a/docs/msbuild/customize-your-build.md
+++ b/docs/msbuild/customize-your-build.md
@@ -20,7 +20,7 @@ MSBuild projects that use the standard build process (importing `Microsoft.Commo
 
 ## Adding arguments to command-line MSBuild invocations for your project
 
-A `Directory.Build.rsp` file in your source directory will be applied to command-line builds of your project. For details, see [MSBuild Response Files](../msbuild/msbuild-response-files.md#directorybuildrsp).
+A `Directory.Build.rsp` file in or above your source directory will be applied to command-line builds of your project. For details, see [MSBuild Response Files](../msbuild/msbuild-response-files.md#directorybuildrsp).
 
 ## Directory.Build.props and Directory.Build.targets
 
--- a/docs/msbuild/customize-your-build.md
+++ b/docs/msbuild/customize-your-build.md
@@ -123,7 +123,7 @@ $(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\{TargetFileName}\ImportAfter\*.t
 
 afterward. This allows installed SDKs to augment the build logic of common project types.
 
-The same directory structure is searched in `$(MSBuildUserExtensionsPath)`, which is the per-user folder `%LOCALAPPDATA%\Microsoft\MSBuild`. Files placed in that folder will be imported for all builds of the corresponding project type run under that user's credentials.
+The same directory structure is searched in `$(MSBuildUserExtensionsPath)`, which is the per-user folder `%LOCALAPPDATA%\Microsoft\MSBuild`. Files placed in that folder will be imported for all builds of the corresponding project type run under that user's credentials. The user extensions can be disabled by setting properties named after the importing file in the pattern `ImportUserLocationsByWildcardBefore{ImportingFileNameWithNoDots}`. For example, setting `ImportUserLocationsByWildcardBeforeMicrosoftCommonProps` to `false` would prevent importing `$(MSBuildUserExtensionsPath)\$(MSBuildToolsVersion)\Imports\Microsoft.Common.props\ImportBefore\*`.
 
 ## Customizing the solution build
 
--- a/docs/msbuild/customize-your-build.md
+++ b/docs/msbuild/customize-your-build.md
@@ -100,7 +100,7 @@ Or more simply: the first *Directory.Build.props* which doesn't import anything,
 
 ## MSBuildProjectExtensionsPath
 
-By default, `Microsoft.Common.props` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.props` and `Microsoft.Common.targets` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.targets`. The default value of `MSBuildProjectExtensionsPath` is `$(BaseIntermediateOutputPath)`, `obj/`. This is the mechanism that NuGet uses to refer to build logic delivered with packages--at restore time, it creates `{project}.nuget.g.props` files that refer to the package contents.
+By default, `Microsoft.Common.props` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.props` and `Microsoft.Common.targets` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.targets`. The default value of `MSBuildProjectExtensionsPath` is `$(BaseIntermediateOutputPath)`, `obj/`. This is the mechanism that NuGet uses to refer to build logic delivered with packages, that is, at restore time, it creates `{project}.nuget.g.props` files that refer to the package contents.
 
 This extensibility mechanism can be disabled by setting the property `ImportProjectExtensionProps` to `false` in a `Directory.Build.props` or before importing `Microsoft.Common.props`.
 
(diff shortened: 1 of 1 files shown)


docs/msbuild/customize-your-build.md
```diff
@@ -15,11 +15,21 @@ ms.workload:
   - "multiple"
 ---
 # Customize your build
+
+MSBuild projects that use the standard build process (importing `Microsoft.Common.props` and `Microsoft.Common.targets`) have several extensibility hooks that can be used to customize your build process.
+
+## Adding arguments to command-line MSBuild invocations for your project
+
+A `Directory.Build.rsp` file in your source directory will be applied to command-line builds of your project. For details, see [MSBuild Response Files](../msbuild/msbuild-response-files.md#directorybuildrsp).
```
jaredpar: Doesn't the searching for rsp files match the searching for props files? Thought that was confirmed in slack.

docs/msbuild/customize-your-build.md
```diff
@@ -86,6 +98,51 @@ A summary of MSBuild's general approach is as follows:
...
 Or more simply: the first *Directory.Build.props* which doesn't import anything, is where MSBuild stops.
 
-## See Also  
- [MSBuild Concepts](../msbuild/msbuild-concepts.md)   
- [MSBuild Reference](../msbuild/msbuild-reference.md)   
+## MSBuildProjectExtensionsPath
+
+By default, `Microsoft.Common.props` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.props` and `Microsoft.Common.targets` imports `$(MSBuildProjectExtensionsPath)$(MSBuildProjectFile).*.targets`. The default value of `MSBuildProjectExtensionsPath` is `$(BaseIntermediateOutputPath)`, `obj/`. This is the mechanism that NuGet uses to refer to build logic delivered with packages--at restore time, it creates `{project}.nuget.g.props` files that refer to the package contents.
+
+This extensibility mechanism can be disabled by setting the property `ImportProjectExtensionProps` to `false` before importing `Microsoft.Common.props`.
```
jaredpar: Can this be disabled via Directory.Build.props? Directory.Build.props is imported after Microsoft.Common.props true but it's ambiguous if it's early enough to disable this property.

docs/msbuild/customize-your-build.md
```diff
@@ -86,6 +98,51 @@ A summary of MSBuild's general approach is as follows:
...
+
+before their contents, and
+
+```
+$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\{TargetFileName}\ImportAfter\*.targets
+```
+
+afterward. This allows installed SDKs to augment the build logic of common project types.
+
+The same directory structure is searched in `$(MSBuildUserExtensionsPath)`, which is the per-user folder `%LOCALAPPDATA%\Microsoft\MSBuild`. Files placed in that folder will be imported for all builds of the corresponding project type run under that user's credentials.
```
jaredpar: Consider listing the MSBuild properties that can be used to override this and when they can be set in the build.

docs/msbuild/customize-your-build.md
```diff
@@ -86,6 +98,51 @@ A summary of MSBuild's general approach is as follows:
...
+
+## Customizing the solution build
+
+> [!IMPORTANT]
+> Customizing the solution build in this way applies only to command-line builds with `MSBuild.exe`. It **does not** apply to builds inside Visual Studio.
+
```
aL3891: Is this also true for directory props/targets files?

docs/msbuild/customize-your-build.md
```diff
@@ -15,11 +15,21 @@ ms.workload:
...
+A `Directory.Build.rsp` file in your source directory will be applied to command-line builds of your project. For details, see [MSBuild Response Files](../msbuild/msbuild-response-files.md#directorybuildrsp).
+
+## Directory.Build.props and Directory.Build.targets
+
 In versions of MSBuild prior to version 15, if you wanted to provide a new, custom property to projects in your solution, you had to manually add a reference to that property to every project file in the solution. Or, you had to define the property in a *.props* file and then explicitly import the *.props* file in every project in the solution, among other things.
 
 However, now you can add a new property to every project in one step by defining it in a single file called *Directory.Build.props* at the root of your repo. When MSBuild runs, *Microsoft.Common.props* searches your directory structure for the *Directory.Build.props* file (and *Microsoft.Common.targets* looks for *Directory.Build.targets*). If it finds one, it imports the property. *Directory.Build.props* is a user-defined file that provides customizations to projects under a directory.
```
aL3891: Perhaps the word repo is confusing here since its actually not really connected to source control, can it be something like "root folder" or "base folder for your code" or something? Come to think of it, what folder is it actually looking at if I run the build from another folder? I suppose its the same folder as the project/solution file is in?

docs/msbuild/customize-your-build.md
```diff
@@ -86,6 +98,51 @@ A summary of MSBuild's general approach is as follows:
 
 Or more simply: the first *Directory.Build.props* which doesn't import anything, is where MSBuild stops.
```
aL3891: "First" here being the "deepest" file, the one firthest down in the directory structure?

docs/msbuild/customize-your-build.md
```diff
@@ -86,6 +98,51 @@ A summary of MSBuild's general approach is as follows:
...
+afterward. This allows installed SDKs to augment the build logic of common project types.
+
+The same directory structure is searched in `$(MSBuildUserExtensionsPath)`, which is the per-user folder `%LOCALAPPDATA%\Microsoft\MSBuild`. Files placed in that folder will be imported for all builds of the corresponding project type run under that user's credentials.
+
+## Customizing the solution build
+
+> [!IMPORTANT]
+> Customizing the solution build in this way applies only to command-line builds with `MSBuild.exe`. It **does not** apply to builds inside Visual Studio.
+
+When MSBuild builds a solution file, it first translates it internally into a project file and then builds that. The generated project file imports `before.{solutionname}.sln.targets` before defining any targets and `after.{solutionname}.targets` after importing targets, including targets installed to the `$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\SolutionFile\ImportBefore` and `$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\SolutionFile\ImportAfter` directories.
```
aL3891: Should this be `after.{solutionname}.sln.targets`?

docs/msbuild/customize-your-build.md
rainersigwald: Correct. The "closest to the searching project"/"first to be found by following this search scheme".

docs/msbuild/customize-your-build.md
rainersigwald: It is _not_. VS builds import d.b.props/targets exactly the same way the command line does.

docs/msbuild/customize-your-build.md
rainersigwald: Yes it should!

docs/msbuild/customize-your-build.md
rainersigwald: Yes, I didn't mean to imply otherwise. Does my update here clarify? I'm trying to avoid duplicating the linked explanation.

docs/msbuild/customize-your-build.md
rainersigwald: Clearer after edit?

docs/msbuild/customize-your-build.md
rainersigwald: 😳 I looked for and **completely** failed to see those properties.

rainersigwald: There are also off switches for the installed wildcard directories; I am deliberately _not_ mentioning them here because they're _very_ likely to break projects. For example, PackageReferences only work because NuGet installs itself to `Microsoft.Common.props\ImportBefore\Microsoft.NuGet.ImportBefore.props`.

docs/msbuild/customize-your-build.md
rainersigwald: Great question! D.B.props is indeed early enough; mentioned that.

docs/msbuild/customize-your-build.md
aL3891: cool!

aL3891:
LGTM!

PR #829, opened 2018-04-19
Add Requirement for only .NET Framework projects

kendrahavens:
Fakes only supports .NET Framework projects. It also cannot support a .NET Framework that targets .NET Standard.

### DIFF
--- a/docs/test/code-generation-compilation-and-naming-conventions-in-microsoft-fakes.md
+++ b/docs/test/code-generation-compilation-and-naming-conventions-in-microsoft-fakes.md
@@ -18,8 +18,8 @@ This article discusses options and issues in Fakes code generation and compilati
 -   Visual Studio Enterprise
 -   A .NET Framework project
 
-[!NOTE]
-.NET Standard projects are not supported.
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


docs/test/code-generation-compilation-and-naming-conventions-in-microsoft-fakes.md
```diff
@@ -16,6 +16,7 @@ This article discusses options and issues in Fakes code generation and compilati
 **Requirements**
 
 -   Visual Studio Enterprise
+-   Only for .NET Framework projects. Currently, there is not support for .NET Framework projects that target .NET Standard.
```
gewarren: @kendrahavens I find this sentence a tiny bit confusing because it says only for .NET Fx, but then not for some .NET Fx. To make it a little clearer, and to match the tense/structure of the first requirement, how about:

- a .NET Framework project

> [!NOTE]
> .NET Standard projects are not supported.

kendrahavens: That sounds good. I'll make this change in the other docs as well.

PR #834, opened 2018-04-19
MSBuild: Version and UriBuilder Property functions

rainersigwald:
I compared the list against

https://github.com/Microsoft/msbuild/blob/e8b480c373eadd9daa975c6e96e4dbbabcd9a4fe/src/Build/Resources/Constants.cs#L408-L436


rainersigwald:
@Mikejo5000

PR #840, opened 2018-04-20
Update error-evaluating-the-function-process-exited.md

jsschultz:
Now that the client app changes are in, modifying the docs to match the final experience.


jsschultz:
@Mikejo5000 can you please take a look? Thanks.

PR #854, opened 2018-04-24
Document MSBuildProjectExtensionsPath and update IntermediateOutputPath

### LINKED ISSUE #853: MSBuild common property (Base)IntermediateOutputPath documentation inaccuracy
Common MSBuild Project Properties [\[GitHub link\]](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/msbuild/common-msbuild-project-properties.md) [\[docs.microsoft.com link\]](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties) appears to make an incorrect statement regarding the `BaseIntermediateOutputPath` property.

The `IntermediateOutputPath` documentation currently states the following (emphasis mine):  
> The full intermediate output path as derived from `BaseIntermediateOutputPath`, if no path is specified. For example, \obj\debug\\. ***If this property is overridden, then setting `BaseIntermediateOutputPath` has no effect.***

However, `BaseIntermediateOutputPath` is used in a few different places regardless of whether or not `IntermediateOutputPath` has been set.

`BaseIntermediateOutputPath` is [used as a default](https://github.com/Microsoft/msbuild/blob/f72bd2eaa37476819ab7647f63d445df8055dab6/src/Tasks/Microsoft.Common.props#L38-L64) for `MSBuildProjectExtensionsPath`, from which `$(MSBuildProjectFile).*.props` is imported if present.

As well, NuGet [uses it as the default](https://github.com/NuGet/NuGet.Client/blob/2b92a51528e0c968466a87449f71c3f2cd704aee/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L573) for `RestoreOutputPath` in some versions, and stores `project.assets.json` and friends in there [by default](https://github.com/NuGet/NuGet.BuildTasks/blob/2c1d484fde94ad7f102f16dd57f3b06fd0402e5f/src/Microsoft.NuGet.Build.Tasks/Microsoft.NuGet.targets#L60).

While NuGet [was recently changed](https://github.com/NuGet/NuGet.Client/blob/deab84db30cbd8cc738cd46ad694aebdbb43a1cf/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L592) to not directly use `BaseIntermediateOutputPath` for the `RestoreOutputPath`, this is not the version that is distributed with the current version of Visual Studio or Mono, and it [looks like a warning is thrown](https://github.com/NuGet/NuGet.Client/blob/deab84db30cbd8cc738cd46ad694aebdbb43a1cf/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L417-L434) if `BaseIntermediateOutputPath` and `MSBuildProjectExtensionsPath` do not match.

I've submitted a pull request containing a potential fix for the documentation in #854

SwooshyCueb:
Closes #853


gewarren:
@Mikejo5000 can you take a look at this PR?
@SwooshyCueb thanks for your contribution and fixing the issue you logged.

PR #938, opened 2018-05-15
a correction of an expression about ZIndex

yilmazdurmaz:
ZIndex value of elements determines their order of appearance and higher values are in the front. The sentence could be changed as "lower values appears behind".


PRMerger10:
@yilmazdurmaz : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #950, opened 2018-05-15
fix: Use relative markdown links in subscription folder

nschonni:
Allows articles to be navigated on GitHub

PR #969, opened 2018-05-17
Fix: piracy/activation.mspx was a 404 page

nschonni:
Sorry, I noticed something off after #950 got merged. Unfortunately the diffs with long text are odd to review, but it looks like I had copy/pasted the link and fixed the pasted version instead of the original.

At the same time, I found that the old link `https://www.microsoft.com/piracy/activation.mspx` was dead , but removing the `.mspx` redirected to https://www.microsoft.com/en-us/piracy/mpa.aspx


PRMerger8:

PRMerger16:

PR #971, opened 2018-05-17
Update foreach-method-map for understandability

lectrician1:
This allows for new programmers to make better connections with the reference above and the example by switching item to value, matching the provided documentation.


PRMerger6:
@lectrician1 : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #977, opened 2018-05-18
Update NUnit Test Generator Extension link for VS 2017

ChrisMaddock:
The NUnit extension was released separately for VS 2017 to the VS 2015 extension. This link points to the VS 2015 version, which is still supported, but I thought it made more sense to point to the VS 2017 version.


PRMerger9:
@ChrisMaddock : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #1013, opened 2018-05-23
Critical differentiation between Self CPU and Total CPU

ryanshripat:
The definitions for *Total CPU (ms)* and *Self CPU (ms)* are exactly the same, making the definition for Self CPU (ms) incorrect. I've added a bit of text from the *Self CPU (%)* definition that makes it correct.


PRMerger12:
@ryanshripat : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #1025, opened 2018-05-24
Update VS for Max product feedback DevCommunity ID

nschonni:
Maybe their is a reason to keep it general?


gewarren:
This might be the correct URL since it's for the VS for Mac topics. @asb3993 and @jdanyow can you comment on this please?

asb3993:
I'm happy to change this to a more specific area. I think it makes more sense!

PR #1026, opened 2018-05-24
Update VS product feedback DevCommunity ID

nschonni:
Maybe their is a reason to keep it general?


gewarren:
@nschonni The original URL redirects to the one you changed it to so I'm not sure this change is the correct thing to do. @jdanyow can you let us know if this URL should be changed as suggested?

nschonni:
No problem, I don't think I noticed the redirect in the first place

PR #1038, opened 2018-05-27
Remove incorrect space from rule name

davkean:
There was additional space in rule name that was incorrect and if pasted in EditorConfig would be incorrect.


PRMerger19:
@davkean : Thanks for your contribution! The author, @kuhlenh, has been notified to review your proposed change.

PR #1049, opened 2018-05-28
M26805: Possible incorrect term: "Django Web Project" vs. "Blank Djan…

carolinacmoravia:
…go Web Project"

Hello, @kraigb,

Translator has reported possible source content issue. 
 
Please review and merge the suggested proposed file change to avoid this error. If you make related fix in another PR, then share your PR number, so we can confirm and close this PR.

Many thanks in advance.


kraigb:
Thanks; you are correct--it should be Web and not App.

kraigb:
#sign-off

PR #1103, opened 2018-06-12
Fix subscription links

nschonni:
Since the Subscription docs are under `/visualstudio/subscriptions/` these should be relative and use the `.md` extension for GH navigation


PRMerger15:

PR #1107, opened 2018-06-12
fix: tfsconfig-cmd links

nschonni:
Not sure if these links changed without a redirect since they don't appear be part of the vsts-docs repo currently


PRMerger19:
@nschonni : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #1142, opened 2018-06-19
Clarify how event handlers differs between Javascript and .NET/C++

### LINKED ISSUE #1141: How event handling differs between .NET/C++ and JavaScript is unclear
I am talking about the last paragraph of https://docs.microsoft.com/en-us/scripting/jswinrt/handling-windows-runtime-events-in-javascript

> In JavaScript, Windows Runtime event arguments are represented as a single event object.

If by “event arguments” you mean “the parameters taken by an event handler”, then yes.

But if by  “**event arg**ument**s**”  you mean “a subclass of [**EventArgs**](https://docs.microsoft.com/dotnet/api/system.eventargs)”, then in VB/C#/C++ too this is a single object!

To be clear, the difference is that in .NET/C++ the *event sender* and the *event data* (the subclass of `EventArgs`) are two separate parameters of the event handler, when in JS both are available through a single parameter (an `Event` object).

(The `EventArgs` name is confusing, it should have been called `EventData`, or simply `Event`. The doc for `EventArgs` says “Represents the base class for classes that contain event data”, but it does not contain *anywhere* the term ‘arguments’! Meanwhile, in .NET like anywhere else, ‘argument’ is generally synonymous with ‘parameter’, see for instance `ArgumentException`. But we have to deal with it.)

> In the following example of an event handler method,

The given example is an event handler *function*. It is not a *method*, it does not belong to any class. (OK, this is a bit pedantic, but still.)

> and the other event arguments. The event arguments are the ones that are documented for each event.

I think it is best to avoid using the term ‘argument’ here, as this can mean two different things (see above). At first when I, a web developer that doesn’t know .NET, read all that, I wondered if we were talking about instantiation of `Event` objects.

PaulCapron:
This is a proposition to fix #1141 (please see the issue discussion for the motivation behind this pull request). I am not very fond of this solution, I am sure it could be improved. But at least it’s a start!

### DIFF
--- a/scripting-docs/jswinrt/handling-windows-runtime-events-in-javascript.md
+++ b/scripting-docs/jswinrt/handling-windows-runtime-events-in-javascript.md
@@ -39,7 +39,7 @@ locator.onpositionchanged =
     };  
 ```  
   
-Another difference between .NET/C++ and JavaScript is the number of parameters taken by an event handler. In .NET/C++, an handler takes two: the event sender, and the event data. In JavaScript, the two are bundled as a single `Event` object. In the following example, the `ev` parameter contains both the sender of the event (the `target` property) and the event data properties (here, just `position`). The event data properties are the ones that are documented for each event.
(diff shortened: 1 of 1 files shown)


PRMerger12:
@PaulCapron : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

PRMerger7:
@PaulCapron : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

Issue #1161, opened 2018-06-25
Document XmlPoke Namespaces parameter


pmisik:
https://github.com/Microsoft/msbuild/issues/3451

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6ef0ef75-5fa0-55e2-415b-e7b3d0c79a56
* Version Independent ID: 5982f1be-d837-2296-9b12-0c353ff1a827
* Content: [XmlPoke Task - Visual Studio](https://docs.microsoft.com/en-us/visualstudio/msbuild/xmlpoke-task)
* Content Source: [docs/msbuild/xmlpoke-task.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/msbuild/xmlpoke-task.md)
* Product: **visual-studio-dev15**
* GitHub Login: @Mikejo5000
* Microsoft Alias: **mikejo**

PR #1162, opened 2018-06-25
Document XmlPoke Namespaces parameter

pmisik:
https://github.com/MicrosoftDocs/visualstudio-docs/issues/1161
https://github.com/Microsoft/msbuild/issues/3451

### DIFF
--- a/docs/msbuild/xmlpoke-task.md
+++ b/docs/msbuild/xmlpoke-task.md
@@ -39,7 +39,7 @@ Sets values as specified by an XPath query into an XML file.
  In addition to having the parameters that are listed in the table, this task inherits parameters from the <xref:Microsoft.Build.Tasks.TaskExtension> class, which itself inherits from the <xref:Microsoft.Build.Utilities.Task> class. For a list of these additional parameters and their descriptions, see [TaskExtension Base Class](../msbuild/taskextension-base-class.md).
 
 ## Example
-Here is sample.xml to modify:
+Here is a sample.xml to modify:
 ```
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
          xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
@@ -48,7 +48,7 @@ Here is sample.xml to modify:
 <mp:PhoneIdentity PhoneProductId="456" PhonePublisherId="0" />
 </Package>
 ```
-If we want to modify `/Package/mp:PhoneIdentity/PhonePublisherId`, then use
+In this example, if you want to modify `/Package/mp:PhoneIdentity/PhonePublisherId`, then use
 ```
 <Project>
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


PRMerger16:
@pmisik : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

docs/msbuild/xmlpoke-task.md
```diff
@@ -29,7 +29,7 @@ Sets values as specified by an XPath query into an XML file.
   
 |Parameter|Description|
 |---------------|-----------------|
-|`Namespaces`|Optional `String` parameter.<br /><br /> Specifies the namespaces for XPath query prefixes.|
+|`Namespaces`|Optional `String` parameter.<br /><br /> Specifies the namespaces for XPath query prefixes. Value specifies list of `Namespace` elements with attributes `Prefix` and `Uri`. The attribute `Prefix` specifies the prefix to associate with the namespace specified in `Uri` attribute. Avoid using empty `Prefix` for default namespace, associate it with some artificial prefix and modify query according to it. This parameter must be XML escaped, so replace &lt; with &amp;lt; and &gt; with &amp;gt;.|
```
rainersigwald: I think I'd say something like

> `Namespaces` is an XML snippet consisting of `Namespace` elements with attributes `Prefix` and `Uri`. The attribute `Prefix` specifies the prefix to associate with the namespace specified in `Uri` attribute. Do not use an empty `Prefix`.

The XML escaping I'm less sure of. It needs to be escaped . . . sometimes. You can store the XML in a property to avoid the escaping, for example

```xml
<Project>
<PropertyGroup>
    <Namespace>
        <Namespace Prefix="dn" Uri="http://schemas.microsoft.com/appx/manifest/foundation/windows10" />
        <Namespace Prefix="mp" Uri="http://schemas.microsoft.com/appx/2014/phone/manifest" />
        <Namespace Prefix="uap" Uri="http://schemas.microsoft.com/appx/manifest/uap/windows10" />
    </Namespace>
</PropertyGroup>

<Target Name="Poke">
  <XmlPoke
    XmlInputPath="Sample.xml"
    Value="MyId"
    Query="/dn:Package/mp:PhoneIdentity/@PhoneProductId"
    Namespaces="$(Namespace)"/>
</Target>
</Project>
```

is equivalent to your example, but more readable.

Mikejo5000: I would go with Rainer's suggestion. If more needs to be said, I'd tend to add it to the Remarks instead: "For the Namespaces parameter, you may need to either store the XML in a property group or include XML escaping. Here's an example with the XML stored in a property group:
<previous code example>

Mikejo5000: And then you could add Rainers example.

PRMerger16:
@pmisik : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

pmisik:
@rainersigwald is right.
Putting namespaces declaration into property is better solution and there is no need to escaping.
I've just tied it and it also works.

PRMerger20:
@pmisik : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

PRMerger14:
@pmisik : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #1246, opened 2018-07-14
Update walkthrough-displaying-light-bulb-suggestions.md

jselbie:
The CreateSuggestedActionsSource should be doing an "or" check instead of an "and" check to validate that both textBuffer and textView are non-null.  The corresponding class that gets constructed, TestSuggestedActionsSource, depends on both of these values to not be null.


PRMerger19:
@jselbie : Thanks for your contribution! The author, @gregvanl, has been notified to review your proposed change.

gregvanl:
#sign-off

PR #1321, opened 2018-08-01
Resolve grammar issues

jpreese:
Couple minor things while looking through the documentation.

As an aside, do you think there's value in creating a sample in the visualstudio docs sample repository for the Bank, and then update the documentation to leverage the sample? 

Right now, the documentation just points to some code on another page: https://docs.microsoft.com/en-us/visualstudio/test/sample-project-for-creating-unit-tests

Once that's in place, I'd like to go through and update some more of the walkthrough documentation to bring it up to date with some best practices if we believe it to be a good idea.


PRMerger7:
@jpreese : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

gewarren:
@jpreese Regarding the aside, can you create a new issue with your sample suggestion, and from there we can determine whether to move forward with it? Since this PR will get closed.

PR #1373, opened 2018-08-09
add a symbol types local and local_function

StevenTCramer:


gewarren:
@StevenTCramer Thank you Steven! @kuhlenh Can you confirm that local and local_function are officially valid applicable types for naming conventions?

gewarren:
@sharwell Can you review this small addition? The other two symbol kinds in the [source code](http://source.roslyn.codeplex.com/#Microsoft.VisualStudio.LanguageServices.Implementation/Options/Style/NamingPreferences/SymbolSpecification/SymbolSpecificationViewModel.cs,45) that aren't in this list are namespace and type parameter. Should those be added too?

sharwell:
> The other two symbol kinds in the source code that aren't in this list are namespace and type parameter. Should those be added too?

Yes, both are implemented and should be added. We have an open issue to add another item but this should be all the ones we have today.

PR #1412, opened 2018-08-21
Remove cloned example code

eko1495:

PR #1416, opened 2018-08-21
Cleanup vsto examples

nschonni:
From looking at #1412 I saw there was a few solution files that had some things that would normally be excluded from source control.
Here are the steps if you'd like to tweak it or run it from the internal repo instead:
1. Add the basic default Visual Studio gitignore https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
1. Commit the gitignore
1. Clear files from index, but leave them on disk `git rm -rf --cached docs/vsto/codesnippet/`
1. Add back the ones that should be tracked git add docs/vsto/codesnippet/
1. Commit the changes

The only part of this cleanup I'm not sure about is the Temporary keys, since they might be required for the bulid (but are likely expired)


nschonni:
It needs to be included here, but I could merge the data with https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/.gitignore if you prefer

I found another one for other codesnippets https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/data-tools/codesnippet/.gitignore

Although this doesn't look like a correct one https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/connected-environment/.gitignore

PR #1422, opened 2018-08-22
Update how-to-add-class-diagrams-to-projects.md

the-wazz:
Updated some text for clarity.

I also added something that I think is needed, but maybe not everywhere:

When referring to a window, I see an approach that is succinct, but not always clear. For ex: "In **Solution Explorer**...", and "In **Class View**...". It's short and sweet but not clear if you don't already know VS. The solution explorer can probably be left as is, but saying, "In Class View", I think, is really vague. So I added some text so it reads, "In the **Class View** window, so it's clearly not a class file but a window. It might not be needed for the solution explorer but it's clearer there too. [Of course, that could mean updating everywhere...]

### DIFF
--- a/docs/ide/class-designer/how-to-add-class-diagrams-to-projects.md
+++ b/docs/ide/class-designer/how-to-add-class-diagrams-to-projects.md
@@ -53,11 +53,11 @@ To add multiple class diagrams, repeat the steps in this procedure.
 
 ## Add a class diagram based on existing types
 
-In **Solution Explorer**, open the class file context menu and then choose **View Class Diagram**.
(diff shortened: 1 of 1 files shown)


PRMerger7:
@the-wazz : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

gewarren:
@the-wazz Thanks for your first contribution to MicrosoftDocs! Our style guidance is not to use the type of UI element after its name, so I'm hesitant to accept this PR. But if you want to change it to include the TIP you mentioned instead, that would be great.

PRMerger7:
@the-wazz : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #1437, opened 2018-08-25
chore: Remove/Add for gitattributes and gitignore

nschonni:
This is probably better done on your internal repo. Here is what I did:
1. `git rm -rf --cached .` removes all files from the git index, but leaves them on the disk
2. `git add .` add back the files looking at the .gitattributes and .gitignore settings
3. Commit the re-encoded files


PRMerger15:
@nschonni : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

ghogen:
@nschonni This is too large a change to accept on the public repo, so maybe enter an issue and describe the advantages of making this change and we can discuss how to move forward. Thanks!

PR #1446, opened 2018-08-27
Fix minor code example discrepancy: csharp_style_expression_bodied_indexers

thisBrian:
Fixed minor discrepancy between true/false indexed `_values` example of csharp_style_expression_bodied_indexers


PRMerger16:
@yourBrian : Thanks for your contribution! The author, @kuhlenh, has been notified to review your proposed change.

PR #1451, opened 2018-08-28
update gitignore to ignore VS local machine (localized) files

eriawan:
Current gitignore needs to be updated because it doesn't have ignored filed for VS specific localized (local machine) files. These ignored files are specific on a user machine, therefore it is safe to ignore.

The files ignored are taken partially from the default gitignore when creating new git repo from VS 2017.


nschonni:
These might be a better template to use if you're going to sync this into the root https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
There is also a few nested `.gitignores` that could be removed if your do that

eriawan:
@nschonni 

> There is also a few nested .gitignores that could be removed if your do that

I agree. I'll look into that too.

gewarren:
@eriawan Thanks for adding these.

Issue #1106, opened 2018-06-12
Incorrect registry key


reednj:
The registry key should be `\HKEY_LOCAL_MACHINE\SOFTWARE\MICROSOFT\.NETFramework\Security\TrustManager\PromptingLevel`

This older version of the article is more accurate: https://msdn.microsoft.com/en-us/library/ee308453.aspx?cs-save-lang=1&cs-lang=csharp#code-snippet-2

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dbdda03d-60e4-adbb-9e80-3cc613c0beb2
* Version Independent ID: 7d6fe3fc-f0c8-3c9b-b331-3fca4417d825
* Content: [How to: Configure the ClickOnce Trust Prompt Behavior - Visual Studio](https://docs.microsoft.com/en-us/visualstudio/deployment/how-to-configure-the-clickonce-trust-prompt-behavior)
* Content Source: [docs/deployment/how-to-configure-the-clickonce-trust-prompt-behavior.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/deployment/how-to-configure-the-clickonce-trust-prompt-behavior.md)
* Product: **visual-studio-dev15**
* GitHub Login: @Mikejo5000
* Microsoft Alias: **mikejo**

Mikejo5000:
@reednj Thank you for pointing it out. I'll take a look.

Mikejo5000:
Thanks!

PR #1455, opened 2018-08-28
typo fixes

csrowell:
It had regedit32 in one place and regedit in another. Also, https://github.com/MicrosoftDocs/visualstudio-docs/issues/1106 only fixed one of the three incorrect registry paths. "Previewing" the changes better shows what's really going on as the raw code looks "wrong" even though that's what you apparently need to get the slash dot (\\.) to display correctly.


ktoliver:
@Mikejo5000 - Would you review the proposed changes? Thanks.

PR #1520, opened 2018-09-13
"NOTE" does not appear to be one of the default tokens (using VS 15.8.4)

jimsvobodacode:
the image should also be updated


PRMerger20:
@jimsvobodacode : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #1703, opened 2018-10-17
Changed "Beginning in 2015"

carlos-quintero:
"Beginning in 2015" is not the correct term, it's  "Beginning with Visual Studio 2015", since it's the product (and not the year) what forces to use SHA256 encryption


PRMerger17:
@carlos-quintero : Thanks for your contribution! The author, @gregvanl, has been notified to review your proposed change.

madskristensen:
👍

PR #1784, opened 2018-10-30
fix error in example for import-group element

zochris:
This fixes an error in the [example](https://docs.microsoft.com/en-us/visualstudio/msbuild/importgroup-element?view=vs-2017#example) for the `ImportGroup` Element.


PRMerger8:
@zochris : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #1873, opened 2018-11-13
Try to make role of devenv.com clearer

drewnoakes:
In my first reading of this it wasn't clear that a `.com` file with the same name was intercepting commands (presumably because `devenv.exe` is a UI application without a console).


PRMerger10:
@drewnoakes : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #1875, opened 2018-11-13
Format example command lines

drewnoakes:
Wrap with backticks, and use lowercase `devenv`.


gewarren:
Thanks Drew! I'll share that with the team.

PR #1886, opened 2018-11-14
Fix error in extending-the-output-window.md

bigrecat:
The IVsOutputWindow interface does not defines methods for showing panes, hiding panes, and manipulating their text. The IVsOutputWindowPane interface defines these methods.


PRMerger15:
@bigrecat : Thanks for your contribution! The author, @gregvanl, has been notified to review your proposed change.

gregvanl:
@bigrecat Thank you for this clarification.
#sign-off

PR #2106, opened 2018-12-26
fix: Remove `~~~` code fence

nschonni:
The three tildas get interpreted as a code fence. The contents are an image note though and don't appear to be a code block


PRMerger6:

PR #2110, opened 2018-12-26
fix: Missing spaces in example web.config XML

nschonni:
spaces where missing between the property names and values.
XML copy also appears to be an old XML language block rather than a functional addition


PRMerger8:
@nschonni : Thanks for your contribution! The author, @ghogen, has been notified to review your proposed change.

PR #2118, opened 2018-12-28
chore: Remove ~~~ code fence for content block

nschonni:
Not sure if the intention was for something like a `---` for a horizontal rule, but the block contents don't appear to be indented to be in a code fence


PRMerger18:

PR #2123, opened 2018-12-28
typo: develpment -> development

nschonni:


jasongroce:
Good catch, thanks @nschonni ! #sign-off

PR #2124, opened 2018-12-28
typo: alows -> allows

nschonni:


nschonni:
Think there is a metadata issue and it should be @corob-msft for the notification

Jak-MS:
@ghogen @corob-msft - I agree with @nschonni that the ms.author metadata attribute seems incorrect in this file. can you update the article in a followup PR (ideally in the private repo)? Thanks!

I'm merging this - the typo fix is good.

PR #2157, opened 2019-01-03
Extend CA2235 rule description to cover ISerializable

drewnoakes:
Per dotnet/roslyn-analyzers#1971, this diagnostic will no longer fire for types implementing `ISerializable`.

//cc: @mavasani


PRMerger19:
@drewnoakes : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #2213, opened 2019-01-15
MSBuild: Remove incorrect Import note

rainersigwald:
This note suggested placing Import statements in a Choose/When
construct, but MSBuild does not appear to have ever supported that.

Removing the entire note, because I believe the conditional-import
restrictions in VS are now relaxed.

### DIFF
--- a/docs/msbuild/import-element-msbuild.md
+++ b/docs/msbuild/import-element-msbuild.md
@@ -67,10 +67,6 @@ Imports the contents of one project file into another project file.
 
  The schema of an imported project is identical to that of a standard project. Although [!INCLUDE[vstecmsbuild](../extensibility/internals/includes/vstecmsbuild_md.md)] may be able to build an imported project, it is unlikely because an imported project typically does not contain information about which properties to set or the order in which to run targets. The imported project depends on the project into which it is imported to provide that information.  
 
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger12:
@rainersigwald : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#hold-off

tmeschter:
@rainersigwald With the _new_ project system/SDK-style projects, I would expect any changes to a project or its imported projects to cause it to be fully and properly re-evaluated.

With the _old_ project system/non-SDK-style projects things are a bit different. Changes to the project file itself will always cause it to be re-evaluated. The issue we still have is that we don't reload _imported_ projects when _they_ change, with a few specific exceptions (notably, the .props/.targets files created during NuGet restore). That may have downstream effects where conditionals in the project file don't evaluate the way you expect them to. Also, not every change to the project will lead to a re-evaluation or new design-time build, so if an `<Import>` is conditioned on a property then we may not pull in (or exclude) the import right away. Instead, that change will only be reflected on the next operation that *does* cause a new design-time build.

All that being said, I think it's reasonable to just delete this section. The statement isn't universally true, and even in the circumstances it is true I'm not sure what a user would do with that knowledge.

Mikejo5000:
#sign-off

rainersigwald:
Thanks all!

PR #2246, opened 2019-01-21
Fix VS Edition Feature Table URL Fragment

AdrianJSClark:
The URL fragment linking to the information about what architecture support there was in each Visual Studio Edition was not correct.


PRMerger15:
@AdrianJSClark : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #2259, opened 2019-01-23
Remove outdated 'Natvis does not work' section

gregg-miskelly:
The natvis page had a note at the top which was from the Visual Studio 2012 days when Concord was only used as the native engine for select scenarios. As of VS 2017, the old native engine doesn't even exist in the product, so this can just be removed.

### DIFF
--- a/docs/debugger/create-custom-views-of-native-objects.md
+++ b/docs/debugger/create-custom-views-of-native-objects.md
@@ -20,10 +20,6 @@ The Visual Studio *Natvis* framework customizes the way native types appear in d
 
 Natvis replaces the *autoexp.dat* file in earlier versions of Visual Studio with XML syntax, better diagnostics, versioning, and multiple file support.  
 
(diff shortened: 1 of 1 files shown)


PRMerger18:
@gregg-miskelly : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
@gregg-miskelly Thanks, Gregg. Do you know if this is likely to affect other scenarios where a feature isn't supported when using mixed mode debugging? Seems like I've seen a few notes about mixed mode debugging limitations, but maybe those don't have anything to do with Concord.

Mikejo5000:
#sign-off

gregg-miskelly:
@Mikejo5000 If we have any other expression evaluator topics that call out interop restrictions - they are probably wrong. Otherwise the restriction is probably true -- there are a lot of unsupported things in interop mode.

PR #2309, opened 2019-02-03
Fix code

lindexi:
The CreatePane return void but the catch return OutputWindowPane that can not compile


PRMerger6:
@lindexi : Thanks for your contribution! The author, @gregvanl, has been notified to review your proposed change.

gregvanl:
Thanks!
#sign-off

PR #2316, opened 2019-02-04
fix: "NOTE" should have exclamation inside brace

nschonni:


GitHubber17:
Verified markdown syntax correction. Thanks!

PR #2310, opened 2019-02-03
Add missing "Organize usings" .editorconfig setting for separating using groups

tiesmaster:
In [15.5](https://github.com/dotnet/roslyn/pull/21087) ["Separate using directive Groups"](https://docs.microsoft.com/en-us/visualstudio/ide/reference/options-text-editor-csharp-advanced?view=vs-2017#using-directives) was added, but the `.editorconfig` setting (`dotnet_separate_import_directive_groups`) was missing. This PR adds that.

PR #2355, opened 2019-02-06
Fix incorrect default value added in #2310

tiesmaster:
@gewarren I was just checking if my addition was already live on the site, but saw I added the default value for it incorrectly 🤦‍♂️ This PR corrects that ;)


PRMerger7:
@tiesmaster : Thanks for your contribution! The author, @kuhlenh, has been notified to review your proposed change.

TrishaMc:
@gewarren - Would you review the proposed changes? Thanks.

PR #2741, opened 2019-02-22
Fix missing

changeworld:


PRMerger7:
@changeworld : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

gewarren:
@changeworld Thank you for suggesting this change. The H3 heading in this case is intentional, otherwise we'd have a single entry "See also" in the in-article TOC.

nschonni:
@gewarren I came across these in some of the other repos. Heading levels shouldn't be skipped for the TOC issue because it doesn't follow HTML semantics and causes issues for screen reader users

PR #2763, opened 2019-02-25
Fix display collapse

changeworld:
Evidence: evi


PRMerger7:
@changeworld : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

ktoliver:
We're reviewing this pull request.

Jak-MS:
@gewarren - looks like almost all the changes are whitespace edits.  The table looks like it renders fine in the GitHub view. I don't think we need to move this to private, but we will if you need to view this in staging.  Thanks.

PR #2834, opened 2019-03-07
fix MSBuild property function IsOSUnixLike

dlech:
IsOSUnixLike is a method, not  a property, so it needs parentheses

`Condition="$([MSBuild]::IsOSUnixLike)"` fails with `error MSB4186: Invalid static method invocation syntax: "[MSBuild]::IsOSUnixLike". Method '[MSBuild]::IsOSUnixLike' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order.`

`Condition="$([MSBuild]::IsOSUnixLike())"` works as expected


PRMerger7:
@dlech : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
@dlech Thanks for the correction!

Mikejo5000:
#sign-off

PR #2855, opened 2019-03-11
Change quick launch use cases to more real life

GKotfis:
`@menu font` - is something that not exists. `@menu project` is more real use case I think.
Same for `@doc font`. I've changed it to `@doc program.cs` - to point to a file.


PRMerger6:
@GKotfis : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #2861, opened 2019-03-12
Update create-an-offline-installation-of-visual-studio.md

MrRodrigoAntunes:
Additional flag was required to download full layout for vs_professional.


PRMerger6:
@MrRodrigoAntunes : Thanks for your contribution! The author, @TerryGLee, has been notified to review your proposed change.

TerryGLee:
@MrRodrigoAntunes We thank you for your input to our documentation! However, after review, we've decided to decline this particular PR. After checking in with the engineering setup team, they've confirmed that both `vs_community.exe   --layout c:\vs2017layout --lang en-US` and `vs_community.exe --layout c:\vs2017layout --lang en-US --all` commands result in identical behavior, so the content is correct as is.
(@tydunkel FYI)

PR #2880, opened 2019-03-15
fix: docs/debugger/includes/remote-debugger-symbols.md

nschonni:
Remove whitespace after YAML frontmater
Doesn't look like this was causing an issue, but sometimes parsers seem to choke when the frontmatter `---` lines contain trailing space


ktoliver:
Merging the typo fix.

@nschonni Thanks!

PR #2894, opened 2019-03-19
Order of AfterTargets/BeforeTargets is incorrect

igor-uzhviev:
The order of AfterTargets and BeforeTargets is incorrect in in the last section "Determine the target build order".
However it is correct in the example above (about Compile, Optimize, Link).

To be more specific AfterTargets means after which targets the current target will be executed, BeforeTargets means before which targets the current target will be executed.
So the order is: AfterTargets -> the current target -> BeforeTargets.

It seems to be it is just a mistype in the documentation.

### DIFF
--- a/docs/msbuild/target-build-order.md
+++ b/docs/msbuild/target-build-order.md
@@ -108,11 +108,11 @@ Targets must be ordered if the input to one target depends on the output of anot
 
 4.  Before a target is executed or skipped, if its `Condition` attribute was absent or did not evaluate to `false`, its `DependsOnTargets` targets are run.
 
-5.  Before a target is executed or skipped, any target that lists it in a `BeforeTargets` attribute is run.
+5.  Before a target is executed or skipped, any target that lists it in a `AfterTargets` attribute is run.
 
 6.  Before a target is executed, its `Inputs` attribute and `Outputs` attribute are compared. If MSBuild determines that any output files are out of date with respect to the corresponding input file or files, then MSBuild executes the target. Otherwise, MSBuild skips the target.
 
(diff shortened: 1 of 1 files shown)


Mikejo5000:
@igor-uzhviev Thanks for working to improve the docs.
You are right about the definitions, it's the phrasing that can be confusing. To use the good example (Optimize, Link) that you cited, step 5 could be restated this way:

Before a target (e.g., "Link") is executed or skipped, any target ("Optimize") that lists it ("Link") in a BeforeTargets attribute is run.

I will try to think of a better phrasing that might avoid confusion. But the existing text is, I believe, technically correct.

PRMerger7:
@igor-uzhviev : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

igor-uzhviev:
> @igor-uzhviev Thanks for working to improve the docs.
> You are right about the definitions, it's the phrasing that can be confusing. To use the good example (Optimize, Link) that you cited, step 5 could be restated this way:
> 
> Before a target (e.g., "Link") is executed or skipped, any target ("Optimize") that lists it ("Link") in a BeforeTargets attribute is run.
> 
> I will try to think of a better phrasing that might avoid confusion. But the existing text is, I believe, technically correct.

Thanks for looking into this. Yes, I agree that the wording is really confusing, especially considering the difference between DependsOnTargets and AfterTargets.
However it would be great if the order in this list would be according to the execution order (e.g. #1 is executed first, then #2, etc.)

Also if you are planning to do some updates in this document it would be nice if you can add some more explanations what the difference between DependsOnTargets and AfterTargets is (something like it is explained here: https://stackoverflow.com/a/2856018/3681541).

Thanks again!

PR #2924, opened 2019-03-23
Update updating-an-existing-application.md

shahabhijeet:
I was confused that I need to add ExcludeAssets="runtime" to every msbuild package reference.
But in this case, you should add it to every MSBuild package reference except Microsoft.Build.Locator

### DIFF
--- a/docs/msbuild/updating-an-existing-application.md
+++ b/docs/msbuild/updating-an-existing-application.md
@@ -44,6 +44,7 @@ For example, you can use this XML:
 <ItemGroup>
   <PackageReference Include="Microsoft.Build" Version="15.1.548" ExcludeAssets="runtime" />
   <PackageReference Include="Microsoft.Build.Utilities" Version="15.1.548" ExcludeAssets="runtime" />
+  <PackageReference Include="Microsoft.Build.Locator" Version="1.1.2"  />
 </ItemGroup>
 ```
 


PRMerger19:
@shahabhijeet : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
@rainersigwald Rainer, this look OK?

docs/msbuild/updating-an-existing-application.md
```diff
@@ -44,6 +44,7 @@ For example, you can use this XML:
 <ItemGroup>
   <PackageReference Include="Microsoft.Build" Version="15.1.548" ExcludeAssets="runtime" />
   <PackageReference Include="Microsoft.Build.Utilities" Version="15.1.548" ExcludeAssets="runtime" />
+  <PackageReference Include="Microsoft.Build.Locator" Version="1.1.2"  />
```
rainersigwald: This is specified separately in the `Add package reference` section below. For step-by-step upgrade instructions, I think it makes sense to do the upgrade in two phases:

* Move to package references
* Move to Locator

But this is definitely confusing if you just come to this document looking for "what does this look like in the end?" which is a totally reasonable thing to do. @Mikejo5000 should we have a separate doc for that use case and establish some cross-links?

ghogen: @Mikejo5000 can you take a look here and recommend the next step?

Mikejo5000: @rainersigwald Rainer, the new non-sequential use case here would be "How to reference MSbuild"? Which would apply equally to MSBuild 15.0 and 16.0.

Mikejo5000:
Opened a PR to make a few minor clarifications

PR #2974, opened 2019-04-02
Including CodeCracker

carloscds:
include CodeCracker analyser.

### DIFF
--- a/docs/ide/csharp-developer-productivity.md
+++ b/docs/ide/csharp-developer-productivity.md
@@ -120,6 +120,7 @@ Several community members have written free extensions that add additional code 
 - [Roslynator](https://marketplace.visualstudio.com/items?itemName=josefpihrt.Roslynator2017)
 - [SonarLint for Visual Studio](https://marketplace.visualstudio.com/items?itemName=SonarSource.SonarLintforVisualStudio2017)
 - [StyleCopAnalyzers](https://www.nuget.org/packages/stylecop.analyzers/)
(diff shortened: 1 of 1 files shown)


PRMerger8:
@carloscds : Thanks for your contribution! The author, @kuhlenh, has been notified to review your proposed change.

gewarren:
@kendrahavens @mavasani Is this okay to include?

kendrahavens:
Hi @carloscds, could you give a brief description of why this change should be included here? Forgive me, I'm not too familiar with the CodeCracker project. There are many analyzer packs in the ecosystem that we *could* link to in our docs, but I don't think the docs are the best discoverability mechanism for the hundreds of options out there.

@jinujoseph @sharwell In case, you have input on this as well.

carloscds:
@kendrahavens CodeCracker it's one of fist project about Roslyn Anayzers. Microsoft already use it in many events and @sharwell know this project. You can seek more about at https://github.com/code-cracker/code-cracker

kendrahavens:
Thanks @carloscds. We can add it then! 

My minor concern was this doc could become overloaded with external links and make discovering the built-in tools (the main purpose of docs.microsoft.com) more difficult. That's definitely not a worry at this stage so we can add the link you suggest.

carloscds:
@kendrahavens Thanks, for me CodeCracker is the same of StyleCopAnalyzers, for example.

sharwell:
@carloscds Yep that was my view as well

PR #3011, opened 2019-04-05
Fix semantic's confusion and add missing pseudovariable

jerryk414:
The documentation for this is a bit confusing, there is a table for Native debugging, C#/Visual Basic debugging, and then Visual Basic debugging again.

Additionally, the pseudovariable '$returnvalue' IS supported in C# (I do not know if it is supported in Visual Basic or Native, but I do know that it is supported in C#.

This is meant to address the confusion, and add the $returnvalue variable to the C# pseudovariable table.

Additionally, there were some discrepancies in semantics that have been normalized:
+  $returnvalue is now lower-case to match the rest of the pseudovariables
+  Use "in the following table" everywhere instead of "in this table"

### DIFF
--- a/docs/debugger/pseudovariables.md
+++ b/docs/debugger/pseudovariables.md
@@ -26,7 +26,7 @@ Pseudovariables are terms used to display certain information in a variable wind
 
 `$handles`
 
- In native code, you can use the pseudovariables shown in this table:
+ In native code, you can use the pseudovariables shown in the following table:
 
 |Pseudovariable|Function|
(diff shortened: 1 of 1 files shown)


PRMerger17:
@jerryk414 : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
@jerryk414 Thanks for the corrections!

Mikejo5000:
#sign-off

PR #3163, opened 2019-04-29
chore: Cleanup redirect files

nschonni:
Using MS Docs authoring pack "Generate Master redirect file".
Manually adjusted for the VS 2015 query string.
Some redirects were relative which it also doesn't handle right now.


PRMerger7:
@nschonni : Thanks for your contribution! The author, @, has been notified to review your proposed change.

ktoliver:
@gewarren - Could you review the proposed changes? Thank you.

PR #3237, opened 2019-05-12
Changed Find the specific object instance in the **Locals** window to…

omarv74:
… Find the specific object instance ...

The section correctly starts with "There are times when you want to observe the behavior of a specific object" then in step 2, needed to apply that change. AFAIK, you cannot find breakpoints in the Locals window, you find specific object instances (variables).


PRMerger6:
@omarv74 : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #3254, opened 2019-05-15
Change project string wrapper character from ` to '

walterlv:
MSBuild project strings should be wrapped with `'` but not `` ` ``.

Although both can be compiled correctly, this article below doesn't mention it, which means it may not be recommended.

- [MSBuild Special Characters - Visual Studio | Microsoft Docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-special-characters?view=vs-2019)


PRMerger15:
@walterlv : Thanks for your contribution! The author, @mikejo5000, has been notified to review your proposed change.

GitHubber17:
Verified correction to syntax. Thank you!

PR #3319, opened 2019-05-26
Global binding Edit.GoToFindCombo overridden

angusgraham:
In VS2019's default mapping Ctrl-D is still bound to Edit.GoToFindCombo in the Global context, but it has no effect that I can see because Ctrl-D is mapped to Edit.Duplicate in the Text Editor context so change this doc to reflect what actually happens.


PRMerger16:
@angusgraham : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

docs/ide/default-keyboard-shortcuts-in-visual-studio.md
```diff
@@ -155,6 +155,7 @@ The sections in the following table include commands that are global in that you
 |Edit.Cut|**Ctrl+X**<br /><br /> or<br /><br /> **Shift+Delete**|
 |Edit.CycleClipboardRing|**Ctrl+Shift+V**<br /><br /> or<br /><br /> **Ctrl+Shift+Ins**|
 |Edit.Delete|**Delete**|
+|Edit.Duplicate|**Ctrl+D**|
```
gewarren: @angusgraham I was not able to get Ctrl+D to do anything in the editor either - were you? It just seems to wait for a second keystroke.

angusgraham: That's odd it does work for me in both VS2019 and VS2017.
If you go to Tools Options Keyboard and put Ctrl-D in the shortcut edit box and then look at what items appear in the "Shortcut currently used by" what do you see?

gewarren: I see, I think this is because you're in the C++ profile. If you choose C# under "Apply the following additional keyboard mapping scheme" you'll see that Ctrl+D disappears as a shortcut for Edit.Duplicate. I'm not sure the best way to handle that in this doc or just how many shortcuts are different between different profiles. I guess for this one you can just put (C++ only) after the shortcut.

angusgraham: I don't see that option. These are the options I have in that dropdown:
(Default)
ReSharper (Visual Studio)
Visual Basic 6
Visual C# 2005
Visual C++ 2
Visual C++ 6
Visual Studio 6
Visual Studio Code

gewarren: Sorry, I meant "Visual C# 2005".

angusgraham: Oh yeah... I don't think you would want have shortcuts from 2005 applied - you don't want to document those any more right?

gewarren: I *think* those correspond to the current C# profile keyboard shortcuts, but I could be wrong. Another way to demonstrate what I'm trying to say is to switch to the C# profile and then look at Tools > Options > Keyboard for Edit.Duplicate. To switch profiles, use the Tools > Import and Export Settings option, choose reset settings, and then choose the Visual C# profile.

angusgraham: I tried what you suggested. I selected "Visual C# 2005" and applied. It said "Keybaord mapping will reset to their defaults. Continue?" and I said "Yes".
But Ctrl-D still works as duplicate! This is in a C# solution or a C++ solution.

As for the meaning of "Visual C# 2005"... Hmm... Except for ReSharper, everything in that list is a boxed product.
For example "Visual C++ 6" is what it was called in 1998 before the moniker "Visual Studio" was invented.
So I assume "Visual C# 2005" means either a boxed product by that name, or maybe the keys were defaulted when Visual Studio 2005 was installed with a C# profile?
I just don't think (C++ Only) is the right thing here, because this key clearly works for me in C# just fine.
On the other hand I have no idea why it works for me and not you.
@kendrahavens do you have some idea what's going on here?

kendrahavens: @gewarren Ctrl+D is mapped to 'Duplicate text' if your cursor is in the Editor under the (Default) Visual Studio keyboard profile. If you have a different window selected Ctrl+D will do nothing. That may be why it did not work for you.
image

The 'C# 2005' keyboard profiles are older then the default ones and I believe we only keep them around to let users retain their muscle memory. We want to document the (Default) options and Ctrl+D is one of them.

Jak-MS:
@gewarren - I'm unable to resolve the merge conflicts in the GitHub web editor UI for some reason. these will need to be resolved, and you should sign off if this is ready for merge. 

thanks.

tagging @angusgraham @kendrahavens

docs/ide/default-keyboard-shortcuts-in-visual-studio.md
angusgraham: If you have telemetry for who's using those I betcha you'd find aside from Resharper and VSCode no one uses those much anymore now it's been ten years.

gewarren:
I switched to the Default profile and now Ctrl+D works. I guess it just doesn't work in the Visual C# profile. I'll clone this repo now so I can resolve the conflict.

gewarren:
@angusgraham I'm not sure why it's now asking you to sign the CLA. Have you previously signed it?

gewarren:
#sign-off

PRMerger17:
@gewarren : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

angusgraham:
> @angusgraham I'm not sure why it's now asking you to sign the CLA. Have you previously signed it?

Yeah that was weird I thought I had. In any case it's def signed now.

PR #3322, opened 2019-05-26
Edit.OpenFile vs File.OpenFile

angusgraham:
It's not obvious how Edit.OpenFile and File.OpenFile are different, so add a note.

Note that Edit.OpenFile is currently broken in VS2019, but set to be fixed in 16.2:
https://developercommunity.visualstudio.com/content/problem/535917/shift-ctrl-g-is-mapped-to-editopenfile-but-isnt-wo.html


PRMerger8:
@angusgraham : Thanks for your contribution! The author, @gewarren, has been notified to review your proposed change.

PR #3410, opened 2019-06-11
Move adding of System and System.Drawing references

pcoleboden:
Add System and System.Drawing references in step 4 with the other references, not when randomly talking about the custommessage parameter

It is better to add all references in one place at the start. It was jarring to go back and add them after creating the wizard class.


PRMerger19:
@pcoleboden : Thanks for your contribution! The author(s) have been notified to review your proposed change.

madskristensen:
#sign-off

PR #3414, opened 2019-06-12
Remove "Lightweight solution load" reference

davkean:
This feature does not exist in VS 2019.


PRMerger6:
@davkean : Thanks for your contribution! The author(s) have been notified to review your proposed change.

gewarren:
@davkean We share these files for the VS 2017 and VS 2019 docs. Does it make sense to also remove this from the VS 2017 docs, or should we just enable that section for the VS 2017 docs?

PR #3423, opened 2019-06-13
Document the selection-modifying commands in editor context

angusgraham:
Expand/contract selection is described on https://devblogs.microsoft.com/visualstudio/improving-your-productivity-in-the-visual-studio-editor/ (tho the shortcut is given as "Shift+Alt++" whereas VS itself gives it as "Shift+Alt+=" which I assume is the correct way to state it.
I don't know if that feature ever worked in C++ editor but as bugged in https://developercommunity.visualstudio.com/idea/605357/expandcontract-selection-does-not-work-for-c.html it does not now so documented that (the bug has been changed to a "feature request" so apparently that's as-designed).
I checked that it works in other editors and it does work in XAML, C# and JSON editors so I presume it works in all editors other than C++.


PRMerger14:
@angusgraham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3499, opened 2019-06-19
Remove any mention of VSWorkspaceSettings.json

esweet431:
Hey Michael and Colin,
Per a conversation with Will and Steven from the AnyCode team, we have decided to remove VSWorkspaceSettings.json from the documentation. The user was never intended to be user-editable - so it's our fault it's included in the first place and has led to some customer confusion. I've tried to clean up this file - let me know if I'm missing something, if there are other files I need to address, or if you have any questions. Thanks for your help!


PRMerger9:
@esweet431 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3543, opened 2019-06-27
Clarify that .edmx files support EF6 only

irontoby:
The link on this page implies that EF Core supports .edmx files. Switch that link to EF6 instead, and add a note that EF Core does not support .edmx files.


PRMerger18:
@irontoby : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3672, opened 2019-07-23
Fix indentation

drewnoakes:
The indentation of three sibling child items in the ParameterGroup was different, making it look like the first and third were parenting the second.

Also the rest of the file uses two spaces for XML indentation, not four.


PRMerger6:
@drewnoakes : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Mikejo5000:
#sign-off

PR #3694, opened 2019-07-29
Proposal to simplify .NET Core support page

richlander:
I got to this page via these release notes: https://github.com/dotnet/core/blob/master/release-notes/2.1/2.1.12/2.1.12.md#net-core-2112-update---july-09-2019

The table was too complicated for me to easily grasp (and I work on the product). I removed the empty rows, and the use of `v` for version. I also wasn't a fan of the `<` symbol. The combination of `-`, `+` and `<` make me think I'm back in school.

### DIFF
--- a/mac/net-core-support.md
+++ b/mac/net-core-support.md
@@ -13,17 +13,13 @@ The following table describes the versions of .NET Core supported by the stable 
 
 .NET Core SDK Version |Visual Studio for Mac 8.1 (Stable) | Visual Studio for Mac 8.2 (Stable) | Visual Studio for Mac 8.3 (Preview) |
 |---------|---------|---------|---------|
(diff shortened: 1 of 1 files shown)


PRMerger13:
@richlander : Thanks for your contribution! The author(s) have been notified to review your proposed change.

sayedihashimi:
@richlander thanks for the updates, can we move this to the visualstudio-docs-pr repo? I made some changes there yesterday. which are not here.
Another benefit of putting it in that repo is that a preview of the file will be rendered. One of the issues that I fixed in my latest commit was getting rid of some hidden characters that were appearing in Safari. WIth this view I cannot validate if those appear or not.

ktoliver:
The updates appear to be reflected in the live article. We'll close this PR. Thanks.

PR #3818, opened 2019-08-25
 Consistent with documentation

noctwolf:


gewarren:
@rpetrusha This is in reference to [this section](https://docs.microsoft.com/en-us/visualstudio/test/live-unit-testing-start?view=vs-2019&tabs=csharp#create-the-solution-and-the-class-library-project) in Get started with Live Unit Testing. Can you take a look?

gewarren:
The description for the method is:

"HasEmbeddedSpaces returns true if a string contains an embedded white-space character; otherwise, it returns false."

But it also returns false if the string is all white-space characters. Is that what your objection is @noctwolf? I guess 'embedded' in this context might mean that there's a white-space character amongst non white-space characters.

PR #3888, opened 2019-09-10
Update walkthrough-creating-and-running-unit-tests-for-windows-store-…

scout208:
Updated to reflect VS2019.

Just updated a couple of names that have been changed in the VS2019 wizard. Someone should update the screenshots with the latest version. I don't know how the old ones were created to add the annotations so I didn't do it myself.

### DIFF
--- a/docs/test/walkthrough-creating-and-running-unit-tests-for-windows-store-apps.md
+++ b/docs/test/walkthrough-creating-and-running-unit-tests-for-windows-store-apps.md
@@ -28,9 +28,9 @@ The following procedures describe the steps to create, run and debug unit tests 
 
      The **New Project** dialog displays.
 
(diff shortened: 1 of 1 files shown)


PRMerger13:

gewarren:
@scout208 Is this change only applicable to VS 2019? If so, we'll need to add separate instructions for those two steps for VS 2017 and 2019 since the file applies to both. We can do that using moniker tags, e.g.:

::: moniker range="vs-2017"
instructions...
::: moniker-end
::: moniker range=">=vs-2019"
instructions...
::: moniker-end

scout208:
@gewarren I'm not sure yet. I haven't tried it in VS2017 but I will take a look and update it accordingly.

gewarren:
@scout208 Thanks for pointing out that these instructions weren't applicable to VS 2019. Since we share this page for VS 2017 too, I've added instructions for both versions in a PR in our private repo. I'm going to close this PR in favor of that one.

PR #3890, opened 2019-09-10
Update ca1507.md

Youssef1313:
The usage of `nameof` really doesn't add any readability in my opinion. It's just good to use it for maintainability, specially, when renaming arguments.

```csharp
throw new ArgumentNullException(nameof(arg));
throw new ArgumentNullException("arg");
```

Both lines are as same readable.


PRMerger15:

PR #3894, opened 2019-09-10
Update unit-testing-visual-csharp-code-in-a-store-app.md

scout208:
The first implementation of `RangeTest` should used `exptected/1000` instead of `ToleranceHelper` because `ToleranceHelper` is not defined until the refactor so it would not build.


PRMerger20:

PR #3895, opened 2019-09-10
Update unit-testing-visual-csharp-code-in-a-store-app.md

scout208:
The refactor of the result implementation was using wrong variable names. I updated them to match the variable names used for this example. I think it was using the .NET Core sample variable names.


PRMerger9:

PR #3914, opened 2019-09-12
Edit.ExpandSelection/Edit.ContractSelection have been fixed again for C++

angusgraham:
These commands were implemented for C++ in VS2017 15.6:
https://developercommunity.visualstudio.com/idea/351165/make-editexpandselection-and-editcontractselection.html
but then regressed at some point after that.
They have been fixed again as of VS 16.3 Preview 2 so can remove the "non-C++ only" note:
https://developercommunity.visualstudio.com/idea/605357/expandcontract-selection-does-not-work-for-c.html


PRMerger6:
@angusgraham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3998, opened 2019-09-26
Changed formatting of some named files/paths

pedro-w:
Some paths were marked up with asterisks which I believe was intended to format them in italics, but actually caused literal asterisks to appear. I replaced with backticks which seems to be consistent with what has been done elsewhere in the file.


PRMerger10:
@pedro-w : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4026, opened 2019-10-01
Remove long-deprecated MSBuild-task parameters

rainersigwald:
@mikejo5000, is it better to remove these entirely or add a "deprecated, do not use" note? I don't know right now exactly when they started being ignored, but it was a very long time ago--I can find out if necessary.


PRMerger14:

Mikejo5000:
If it's a long time ago, I think we're OK to just remove it.

Mikejo5000:
#sign-off

PR #4043, opened 2019-10-02
Update vs-azure-tools-diagnostics-for-cloud-services-and-virtual-mach…

dcbrown16:
John Kemnez provided this information in internal verification:
> You can specify a smaller amount than 4 GB. The absolute lowest is 50 MB, but there’s a high risk of not uploading data if that is what you set (if more than 50 MB needs to be buffered).


PRMerger18:
@dcbrown16 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen:
#sign-off

PR #4040, opened 2019-10-02
MSBuild: UsingTask is first-one-wins

rainersigwald:
This came up while debugging something with @vatsan-madhavan.

PR #4044, opened 2019-10-03
renaming files and updating security warnings page

eddynaka:
- renaming files
- adding more rules to page based on menu
- updating security warnings md to new urls


jborsecnik:
@gewarren, can you review this PR? Thanks.

docs/msbuild/usingtask-element-msbuild.md
```diff
@@ -34,6 +34,9 @@ Maps the task that is referenced in a [Task](../msbuild/task-element-msbuild.md)
     Condition="'String A'=='String B'" />
 ```
 
+> [!NOTE]
+> Unlike properties and items, the *first* `UsingTask` element that applies to a `TaskName` will be used; to override tasks you must define a new `UsingTask` *before* the existing one.
```
gewarren: @Mikejo5000 or @rainersigwald  Can you review this change in this MSBuild file?

gewarren:
Thank you @eddynaka. Whenever a file is renamed, a redirect needs to be put in place in the [redirection file](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/.openpublishing.redirection.json) so people don't start getting 404s for the old URL. But I also want to first okay this with @mavasani and @dotpaul who have in-product links to these CA rule pages (e.g. in the Visual Studio Error List).

eddynaka:
@gewarren , should i update that file as well? The example below, is that right?
```json
{
    "source_path": "docs/code-quality/ca2100-review-sql-queries-for-security-vulnerabilities.md",
    "redirect_url": "/visualstudio/code-quality/ca2100.md",
    "redirect_document_id": false
},

**Comment by gewarren** [comment]
That's almost correct. It should be like:

```
{
    "source_path": "docs/code-quality/ca2100-review-sql-queries-for-security-vulnerabilities.md",
    "redirect_url": "/visualstudio/code-quality/ca2100",
    "redirect_document_id": true
},
```

I'd like to get input from Manish and Paul before we move forward with bulk renames, however, to make sure they're on board with it.

dotpaul:
If there are redirects for the old URLs, I think the renames are fine.  I wouldn't mind updating the hardcoded URLs in Roslyn analyzers.  @mavasani, sound okay to you?

If we do go down this path, also note there are links within some of the docs, so it might be worth searching for the regex "\(ca\d{4}-"

mavasani:
> I'd like to get input from Manish and Paul before we move forward with bulk renames, however, to make sure they're on board with it.

Oh wait, this will likely break the help link navigation from error list to these doc pages for all our released analyzer packages.

mavasani:
> If there are redirects for the old URLs, I think the renames are fine

Ah, if we are doing the redirects for old ones, it seems fine to me too.

mavasani:
> I wouldn't mind updating the hardcoded URLs in Roslyn analyzers. @mavasani, sound okay to you?

Yes, that should hopefully be easy to do and also make the help links lot more consistent. Overall, I like the change of just using `CAxxxx.md` as the file names for doc files.

docs/msbuild/usingtask-element-msbuild.md
rainersigwald: That looks like it should have come from MicrosoftDocs/visualstudio-docs#4040; it looks ok, but I don't understand why it's showing up here.

gewarren:
Can we also get an SEO OK from @TerryGLee or @cjgronlund? I know that file name (which becomes the URL) plays a role in SEO (findability), however, if we have the diagnostic message in the H1 (heading) and metadata title, I think we should be okay.

Here's an [example topic](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1000-do-not-declare-static-members-on-generic-types?view=vs-2019) - we're proposing to reduce the file name to just the rule ID, i.e. `ca1000.md` instead of `ca1000-do-not-declare-static-members-on-generic-types.md`.

gewarren:
It sounds like we have SEO buyoff to make this change. Also, a lot of traffic to these pages is probably direct traffic (i.e. clicking a link instead of searching).

gewarren:
@eddynaka We can close this PR in the public repo, correct?

eddynaka:
@gewarren yes! just closed! thanks

PR #4072, opened 2019-10-08
Add a step to add UICulture element in project file for 'Localize menu commands'

kaotoby:
When setting `NeutralResourcesLanguage` to `UltimateResourceFallbackLocation.Satellite`, it's required to set a default UICulture if your VSIX conducts WPF and XAML.
If you don't do that, it'll throw a `System.IO.IOException` like below in runtime due to cannot locate the culturalized .xaml resource.
> System.IO.IOException: 'Cannot locate resource 'dialogwindow.xaml'.'


madskristensen:
#signoff

PR #4446, opened 2019-12-06
Added ANCMPreConfiguredForIIS

jcoutch:
This change is to add `ANCMPreConfiguredForIIS` to the list of common MSBuild project properties, which was introduced in Visual Studio/MSBuild 16.3:
https://github.com/aspnet/websdk/issues/564#issuecomment-532900476


PRMerger6:
@jcoutch : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Mikejo5000:
@rainersigwald Rainer, should this property be documented here, since it is specific to IIS?

rainersigwald:
@vijayrkn do you think this should be documented here? Is there a docs page with other ASP.NET-specific properties that would be a better home for it?

vijayrkn:
A better place for this would be in the ASP.NET docs. @bradygaster - Do you know what would be a good place to get the ASP.NET IIS integration properties documented?

/cc @BillHiebert

bradygaster:
I would initially say one of the articles below this one in our TOC on docs. https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/?view=aspnetcore-3.1

ghogen:
Closing, per previous discussions.

PR #4567, opened 2019-12-30
Fix see also section

Youssef1313:
All links were appearing in a single line and looks very bad:

image


PRMerger8:

PR #4608, opened 2020-01-08
Update common-msbuild-project-properties.md

ocallesp:
Adding OutDir property for msbuild.
OutputPath is for legacy. We should use OutDir instead.
https://github.com/Microsoft/msbuild/issues/87


PRMerger19:
@ocallesp : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ocallesp:
Adding this msbuild property in code 
PR
https://github.com/dotnet/project-system/pull/5773

Chatina73:
What are my sources?

ghogen:
It might not be as simple as recommending OutDir only - but it should definitely be added to the documented properties.
#sign-off

PR #4671, opened 2020-01-19
The 'data' variable isn't being used in the sample

ferdeen:
The 'data' variable isn't being used in the sample.  Not sure if this needs to be removed or append it to the Console.WriteLine statements in InstanceMethod() method.

This isn't a code change, but highlighting variables not being used.  Either use them or remove as it could cause confusion for the reader.


PRMerger12:
@ferdeen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Mikejo5000:
Updating doc

Mikejo5000:
Fixes are in master, should be live tomorrow

PR #4754, opened 2020-02-05
Update whats-new-visual-studio-2019.md

bergano65:
isn't clear why workflow is cloud first. correct me if I'm wrong


PRMerger12:
@bergano65 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TerryGLee:
@bergano65 Thanks for the inquiry. While this isn't a true pull request, the answer is that we consider the workflow to be cloud-first because of the Git integration. However, if that's confusing, we'll change the terminology to read "Git-first" instead.

Meanwhile, we'll close this PR in that it's not truly a PR but more of a [GitHub Issue](https://github.com/MicrosoftDocs/visualstudio-docs/issues) instead. (And also, we don't add comments to our Markdown, as a practice.)

PR #4768, opened 2020-02-07
AddDeleteInputFile

shahabhijeet:
Adding a note for scenario when adding new files or deleting existing files does not update output.


PRMerger13:
@shahabhijeet : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen:
@shahabhijeet @rainersigwald 
Let's close this one, but internally we will add a Note.  I think the wording has to very clear.

ghogen:
Closing - thanks for pointing out this issue @shahabhijeet

shahabhijeet:
Would you tell me what you mean by "internally we will add a Note", does that mean that note will not be reflected on the public docs?

PR #4809, opened 2020-02-18
Fix incorrect description for Unity Api Compatibility Level

kiyoaki:


PRMerger18:
@kiyoaki : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@therealjohn 

Hi, could you review this pull request? Thanks.

therealjohn:
@kiyoaki This doesn't seem right to me. Can you explain?

kiyoaki:
> .NET 4.x. This profile provides access to the latest .NET 4 API. It includes all of the code available in the .NET Framework class libraries and supports .NET Standard 2.0 profiles as well. Use the .NET 4.x profile if your project requires part of the API not included in the .NET Standard 2.0 profile. However, some parts of this API may not be supported on all of Unity's platforms.

Unity .NET Standard 2.0 profile is subset of .NET 4.x. profile'.

Correct:
For example, if you're using the .NET Standard 2.0 profile and want to use `HttpClient`, you must add an assembly reference for System.Net.Http.dll.

Incorrect:
For example, if you're using the .NET 4.x profile and want to use `HttpClient`, you must add an assembly reference for System.Net.Http.dll.

therealjohn:
@kiyoaki 

Look at https://docs.unity3d.com/Manual/dotnetProfileAssemblies.html, it seems that this is not correct. 

If you're using the .NET Standard 2.0 profile, and need more, you would switch to .NET 4.x. With that profile, you can reference other assemblies when necessary.

PR #4905, opened 2020-03-07
Update walkthrough-creating-an-msbuild-project-file-from-scratch.md

KinsonDigital:
Fix explanation of the 32 or 64 bit OS based on the ProgramFiles path.


PRMerger7:
@KinsonDigital : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen:
@KinsonDigital On my 64-bit system, this is indeed in "C:\Program Files (x86)" - that's because on a 64-bit system, the Program Files (x86) folder contains 32-bit-compiled applications that run on the 64-bit system, whereas "C:\Program Files" contains 64-bit natively compiled applications.

Issue #4908, opened 2020-03-08
Add information about fixing the container's ports when debugging


rodolphocastro:
The documentation doesn't show how we can change `launchSettings.json` in order to have fixed ports when we are running the container using Visual Studio.

The following stackoverflow question pointed me in the right direction:  https://stackoverflow.com/a/59582605/5342210

Also, when checking `launchSettings.json` with Intelissense there's not a single hint about those mappings as well.

```json
{
"Docker": {
      "commandName": "Docker",
      "launchBrowser": true,
      "launchUrl": "{Scheme}://{ServiceHost}:{ServicePort}/",
      "publishAllPorts": true,
      "useSSL": true,
      "httpPort": 5000,  // This will make sure your container runs on host:5000 when using Http
      "sslPort":  5001 // This will make sure your container runs on host:5001 when using https
    }
}
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cc5c3c25-aa1b-e4ce-dc4f-13270251f462
* Version Independent ID: 20629bb8-5bed-5886-e931-df8dccd1e71f
* Content: [Visual Studio Container Tools with ASP.NET - Visual Studio](https://docs.microsoft.com/en-us/visualstudio/containers/container-tools?view=vs-2019#feedback)
* Content Source: [docs/containers/container-tools.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/containers/container-tools.md)
* Product: **visual-studio-windows**
* Technology: **vs-azure**
* GitHub Login: @ghogen
* Microsoft Alias: **ghogen**

rodolphocastro:
Just noticed there's already some documentation about this on [docs subpage](https://docs.microsoft.com/en-us/visualstudio/containers/container-launch-settings?view=vs-2019), but given how common it is to need a fixed port (such as when you have an auth callback to a SwaggerUi) I belive it deserved at least a mention on the quickstart

rodolphocastro:
The pull request #4909 proposes changes to the current Container Quickstart docs

msebolt:
@rodolphocastro Thank you for your contribution!

Assigning this to the author for further review.

ghogen:
@rodolphocastro Thanks, looking into this now.

PR #4909, opened 2020-03-08
Adds information about launchSettings to Container Quickstart

rodolphocastro:
This commit adds a new section "Changing launch settings" which makes it clear that one can change their `launchsettings.json` in order to customize how their containers are executed within Visual Studio. This section also points to the current page about the LaunchSettings Schema for Docker images

Related: #4908


PRMerger12:
@rodolphocastro : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ShannonLeavitt:
@ghogen 

If you decide to accept this contribution, we'll remove the locale from the URL.

ghogen:
In discussion with the engineering team, we decided not to put this section in this article.  We could consider it in another article, though.

PR #4935, opened 2020-03-13
Explain need for Locator API call

Forgind:
Clarify rationale for separating call to Locator API from method using types

### DIFF
--- a/docs/msbuild/updating-an-existing-application.md
+++ b/docs/msbuild/updating-an-existing-application.md
@@ -79,7 +79,33 @@ Do not specify `ExcludeAssets=runtime` for the Microsoft.Build.Locator package.
 
 ### Register instance before calling MSBuild
 
-Add a call to the Locator API before calling any method that uses MSBuild. This is important because the just-in-time (JIT) compiler needs to understand the types of the objects it compiles, and when trying to build a project using a specific version of MSBuild, it has to inspect that version before it compiles the code in case those types have changed. Since the JIT compiler works at a method-level granularity, if the call to the Locator API is in the same method as the first use of an MSBuild type (even if it precedes it, and the code involving the type has not yet been executed), the JIT compiler will not be able to JIT the whole method, causing an error. Some types that require the call to the MSBuildLocator API to have already been called include ProjectRootElement, ProjectInstance, and BuildManager.
+> [!IMPORTANT]
+> You cannot reference any MSBuild types (from the `Microsoft.Build` namespace) in the method that calls MSBuildLocator. For example, you cannot do this:
+>
+> ```csharp
+> void ThisWillFail()
+> {
+>     MSBuildLocator.RegisterDefaults();
+>     Project p = new Project(SomePath); // Could be any MSBuild type
+>     // Code that uses the MSBuild type
+> }
+> ```
+>
+> Instead, you must do this:
+>
+> ```csharp
+> void MethodThatDoesNotDirectlyCallMSBuild()
+> {
+>     MSBuildLocator.RegisterDefaults();
+>     MethodThatCallsMSBuild();
+> }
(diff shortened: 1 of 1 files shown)


docs/msbuild/updating-an-existing-application.md
```diff
@@ -79,7 +79,7 @@ Do not specify `ExcludeAssets=runtime` for the Microsoft.Build.Locator package.
 
 ### Register instance before calling MSBuild
 
-Add a call to the Locator API before calling any method that uses MSBuild.
+Add a call to the Locator API before calling any method that uses MSBuild. This is important because the just-in-time (JIT) compiler needs to understand the types of the objects it compiles, and when trying to build a project using a specific version of MSBuild, it has to inspect that version before it compiles the code in case those types have changed. Since the JIT compiler works at a method-level granularity, if the call to the Locator API is in the same method as the first use of an MSBuild type (even if it precedes it, and the code involving the type has not yet been executed), the JIT compiler will not be able to JIT the whole method, causing an error. Some types that require the call to the MSBuildLocator API to have already been called include ProjectRootElement, ProjectInstance, and BuildManager.
```
rainersigwald: I think this is overly detailed and would be better off with a good and bad example. Writing one up.

Forgind: Perhaps it's just because I think this sort of thing is really cool, but I would appreciate that level of detail if for no other reason that I'd wonder why Microsoft would put in arbitrary requirements, and this would clarify that. That being said, an example would help, and you could certainly look at it faster.

PRMerger12:
@Forgind : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/msbuild/updating-an-existing-application.md
rainersigwald:
```suggestion
> [!IMPORTANT]
> You cannot use any MSBuild types (from the `Microsoft.Build` namespace) in the method that calls MSBuildLocator. For example, you cannot do this:
>
> ```csharp
> void ThisWillFail()
> {
>     MSBuildLocator.RegisterDefaults();
>     Project p = new Project(SomePath);
>     p.Build();
> }
> ```
>
> Instead, you must do this:
>
> ```csharp
> void MethodThatDoesNotDirectlyCallMSBuild()
> {
>     MSBuildLocator.RegisterDefaults();
>     MethodThatCallsMSBuild();
> }
> 
> void MethodThatCallsMSBuild()
> {
>     Project p = new Project(SomePath);
>     p.Build();
> }
> ```
```

Forgind: Can we change `use` to `reference` since a user also cannot create an MSBuild type even without using it.

Also, I would prefer stopping at `Project p = new Project(SomePath);` since that would already fail, and it shows the minimum required to fail. Adding `p.Build();` makes it a more realistic scenario, but it could conceivably lead to confusion if someone thinks the problem is building rather than referring to a type at all. That being said, being realistic is good, so I could see it both ways.

ghogen: @rainersignwald Can you follow-up on @Forgind's question here?  I think a code comment could help address the confusion, rather than changing the code to remove p.Build().

rainersigwald: Saying "reference" instead of "build" seems fine. Edited my suggestion to make that easy to apply. replaced `p.Build` with a comment.

PRMerger12:
@Forgind : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/msbuild/updating-an-existing-application.md
Forgind: I like it!

PR #5040, opened 2020-03-30
We don't GAC MSBuild assemblies

Forgind:
We shouldn't tell customers that MSBuild is in the GAC. They put it there anyway, and it causes problems.


PRMerger18:
@Forgind : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5066, opened 2020-04-03
.natvisreload must be run in the Immediate Window

htpiv:
The page currently says to reevaluate the .natvisreaload command in the Watch window, which, at least when I tried it on VS Enterprise 16.5.2, doesn't do anything. The command should be run in the Immediate window.


Mikejo5000:
#sign-off

PR #5072, opened 2020-04-05
Update create-new-project.md

bergano65:
matter of style but probably emphasizing project platform and type will make  text more readable

### DIFF
--- a/docs/ide/create-new-project.md
+++ b/docs/ide/create-new-project.md
@@ -61,7 +61,7 @@ If the Visual Studio development environment is already open, you can create a n
 
 On the **Create a new project** page, a list of your recently selected templates appears on the left. The templates are sorted by *most recently used*.
 
-If you're not selecting from the recently used templates, you can filter all available project templates by **Language** (for example, C# or C++), **Platform** (for example, Windows or Azure), and **Project type** (for example, Desktop or Web). You can also enter search text into the search box to further filter the templates, for example, **asp.net**.
+If you're not selecting from the recently used templates, you can filter all available project templates by **Language** (for example, C# or C++), **Platform** (for example, **Windows** or **Azure**), and **Project type** (for example, **Desktop** or **Web**). You can also enter search text into the search box to further filter the templates, for example, **asp.net**.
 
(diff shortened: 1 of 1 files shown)


PRMerger19:
@bergano65 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TerryGLee:
#sign-off

TerryGLee:
@bergano65 I accepted this PR but please know that our internal style is to only bold those UI items that we specifically ask a user to interact with, which was why they weren't bolded initially, as they were suggestions and not specific directions.

bergano65:
> @bergano65 I accepted this PR but please know that our internal style is to only bold those UI items that we specifically ask a user to interact with, which was why they weren't bolded initially, as they were suggestions and not specific directions.

thanks, I'll  use this policy, are they listed somewhere?  probably use italics  for separating from ordinary text and interacting topics also

TerryGLee:
@bergano65 In the Docs folder for this repo, there is the following file, https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/CONTRIBUTING.md, which includes a link to our style guide. For information specific to how to format text instructions, see https://docs.microsoft.com/en-us/style-guide/procedures-instructions/formatting-text-in-instructions (and also the links that are at the top of this page).

As you might note, the guide isn't specific about "suggestions" in a procedure. Which is why our default is, if it isn't directly something you are asked to interact with in UI, then *don't* use bold formatting. The idea here is that people are used to seeing that the things they should "click" are marked in bold. But if we overuse bold for UI elements that don't require a user to directly interact with, then we oversaturate our instructions and risk people missing key elements of our instructions in the "noise".

bergano65:
ok, thanks if like this probably not use my changes?

TerryGLee:
#hold-off

@bergano65 Thanks for the suggestion. I hesitated approving the PR because we appreciate input from our customers and I didn't want to discourage anyone from submitting a PR. But you're right, in this case, we probably shouldn't process the PR. So, I will close this instead. But thanks again for creating the PR, and we hope to see future input from you, going forward!

PR #5175, opened 2020-04-23
MSBuild: ToolTaskExtension snippet

rainersigwald:
People often erroneously think that tasks don't expose arguments like `EnvironmentVariables`, because the (code) structure of the task implementation inherits public properties from `ToolTask` or `ToolTaskExtension`. These were documented, but only as a link in the remarks section.

I'm converting the existing `ToolTaskExtension` argument list to a snippet and including it where it was linked previously.


ghogen:
The include should be in the includes folder and I'd leave the base class article.  You could include the include in the base class topic.  Also, the metadata for the include has specific requirements. See https://review.docs.microsoft.com/en-us/help/contribute/metadata-attributes?branch=master#metadata-for-included-files.

PRMerger13:

rainersigwald:
> @rainersigwald Rainer Sigwald FTE I can take this into the private repo and fix it up in a way that follows guidelines.

That would be great, thanks. Happy to review!

PR #5261, opened 2020-05-08
Fix image from when 'Simplify interpolation' confused left-pad with left-align

jnm2:
See https://github.com/dotnet/roslyn/issues/41826#issuecomment-590287078 for reference. @mikadumont updated the screenshot at https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes#simplify-string-interpolation. I just noticed that https://docs.microsoft.com/en-us/visualstudio/ide/reference/simplify-string-interpolation?view=vs-2019 is still showing the bad image.

So sorry to have made the mistake!


mikadumont:
#sign-off

PRMerger10:
@mikadumont: I'm sorry - only the author of this article, @jillre, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #5380, opened 2020-05-26
Fix screenshot for import completion

genlu:
Update the screenshot for import completion to actually show the feature :)

before:
image

after:
image

You can check what it looks like in the doc from my branch here:
https://github.com/genlu/visualstudio-docs/blob/fixImportCompletion/docs/ide/reference/intellisense-completion-unimported-types.md


genlu:
FYI @mikadumont

PRMerger18:
@mikadumont, you can only sign off in the main PR conversation, not in review comments.

mikadumont:
#sign-off

PR #5458, opened 2020-06-10
Fix inappropriate info

briuk:
Fix incorrect info about Feedback buttons **label** and **position**.
Change inappropriate **link** for article about installing VS SDK.


PRMerger13:
@briuk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5544, opened 2020-07-01
Fix a typo

Youssef1313:
Fixes https://github.com/dotnet/roslyn-analyzers/issues/3805


PRMerger16:

docs/code-quality/ca2007.md
```diff
@@ -41,7 +41,7 @@ To fix violations, call <xref:System.Threading.Tasks.Task.ConfigureAwait%2A> on
 
 ## When to suppress warnings
 
-You can suppress this warning if you know that the consumer is not a graphical user interface (GUI) app or if the consumer does not have a <xref:System.Threading.SynchronizationContext>.
+You can suppress this warning if you know that the consumer is a graphical user interface (GUI) app or if the consumer does not have a <xref:System.Threading.SynchronizationContext>.
```
Youssef1313, line 44: @mavasani, I'm not sure about the other part of the sentence related to the SynchronizationContext

mavasani: Tagging @stephentoub. https://devblogs.microsoft.com/dotnet/configureawait-faq/ for reference.

stephentoub: Thanks.  I'd prefer to see it more like this:
```suggestion
This warning is intended for libraries, where the code may be executed in arbitrary environments and where code shouldn't make assumptions about the environment or how the caller of the method may be invoking or waiting on it.  It is generally appropriate to suppress / disable the warning entirely over projects that represent application code rather than library code; in fact, running this analyzer on application code (e.g. button click event handlers in a Windows Forms project) is likely to lead to the wrong actions being taken.

You can suppress this warning in any situation where either the continuation should be scheduled back to the original context or where there is no such context in place.  For example, when writing code in a button click event handler in a Windows Forms application, in general the continuation from an await should run on the UI thread, and thus the default behavior of scheduling the continuation back to the originating context is desirable.  As another example, when writing code in an ASP.NET Core application, by default there is no SynchronizationContext or TaskScheduler, such that a `ConfigureAwait` wouldn't actually change any behavior.
```

@mavasani, is CA2007 enabled by default in the new SDK analyzers library?  If yes, we should change that.

PRMerger7:

PRMerger8:

docs/code-quality/ca2007.md
Youssef1313: @stephentoub Thanks. It's much more detailed now. I've applied the suggestion and added a few small tweaks (mention WPF too, added xref to class names, replaced e.g. with for example based on [this guidelines](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg).

One question got to my mind now. What if I'm writing a class library that's intended to be used in **my own application only**, and I can **guarantee** that it will **always** be called asynchronously. Does it matter in that case whether I use ConfigureAwait(false) in my class library?

mavasani:
> @mavasani, is CA2007 enabled by default in the new SDK analyzers library? If yes, we should change that.

No, it is disabled. I am not sure if the comment is still valid or not.
https://github.com/dotnet/roslyn-analyzers/blob/0cbdd68e14722e911d50eee64ee95823b3ffa09c/src/NetAnalyzers/Core/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotDirectlyAwaitATask.cs#L30

docs/code-quality/ca2007.md
stephentoub: > Does it matter in that case whether I use ConfigureAwait(false) in my class library?

No, at that point it's basically application code.

docs/code-quality/ca2007.md
```diff
@@ -41,7 +41,9 @@ To fix violations, call <xref:System.Threading.Tasks.Task.ConfigureAwait%2A> on
 
 ## When to suppress warnings
 
-You can suppress this warning if you know that the consumer is not a graphical user interface (GUI) app or if the consumer does not have a <xref:System.Threading.SynchronizationContext>.
+This warning is intended for libraries, where the code may be executed in arbitrary environments and where code shouldn't make assumptions about the environment or how the caller of the method may be invoking or waiting on it. It is generally appropriate to suppress / disable the warning entirely over projects that represent application code rather than library code; in fact, running this analyzer on application code (for example, button click event handlers in a WinForms/WPF project) is likely to lead to the wrong actions being taken.
```
Mikejo5000, line 44:
```suggestion
This warning is intended for libraries, where the code may be executed in arbitrary environments and where code shouldn't make assumptions about the environment or how the caller of the method may be invoking or waiting on it. It is generally appropriate to suppress the warning entirely for projects that represent application code rather than library code; in fact, running this analyzer on application code (for example, button click event handlers in a WinForms or WPF project) is likely to lead to the wrong actions being taken.
```

docs/code-quality/ca2007.md
```diff
@@ -41,7 +41,9 @@ To fix violations, call <xref:System.Threading.Tasks.Task.ConfigureAwait%2A> on
...
+
+You can suppress this warning in any situation where either the continuation should be scheduled back to the original context or where there is no such context in place. For example, when writing code in a button click event handler in a WinForms/WPF application, in general the continuation from an await should run on the UI thread, and thus the default behavior of scheduling the continuation back to the originating context is desirable. As another example, when writing code in an ASP.NET Core application, by default there is no <xref:System.Threading.SynchronizationContext> or <xref:System.Threading.Tasks.TaskScheduler>, such that a `ConfigureAwait` wouldn't actually change any behavior.
```
Mikejo5000, line 46:
```suggestion
You can suppress this warning in any situation where either the continuation should be scheduled back to the original context or where there is no such context in place. For example, when writing code in a button click event handler in a WinForms or WPF application, in general the continuation from an await should run on the UI thread, and thus the default behavior of scheduling the continuation back to the originating context is desirable. As another example, when writing code in an ASP.NET Core application, by default there is no <xref:System.Threading.SynchronizationContext> or <xref:System.Threading.Tasks.TaskScheduler>, for which reason a `ConfigureAwait` wouldn't actually change any behavior.
```

PRMerger14:

Youssef1313:
@Mikejo5000 Is there anything left for this PR?

Mikejo5000:
#sign-off

Issue #5664, opened 2020-08-03
Microsoft Doc Accessibility Workshop for Write the Docs, August 9, 2020


tedhudek:
# Microsoft Doc Accessibility Workshop for Write the Docs, August 9, 2020

We're so glad you could join us today!

We will be bringing the most frequently visited docs in the Visual Studio documentation repo into alignment with accessibility recommendations in the Microsoft Style Guide.  By doing so, we'll set a precedent for the community, as well as Microsoft's writers.

Specifically, we're going to update occurrences of the word click, according to the following style guide recommendations:

[Describing interactions with UI](https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui)

Here's how to proceed:

1. Look at [vs-filenames-with-click](https://gist.github.com/tedhudek/fb0b8117a0006002f77e6c6721f120c6).
2. In the list, pick a block of files to work on.
3. Look in the discussion in this issue to ensure that no one else is already working on that block. Add a comment that you are starting work on it.
4. Either work in the browser (hint: use the `t` keyboard shortcut on GitHub to get to the file picker, where you can quickly find files) or `git clone https://github.com/MicrosoftDocs/visualstudio-docs` and work locally.
5. Open one pull request to contain all your changes. In the PR, add `#5664` in the conversation to link to this issue.
6. Add the `Write-the-Docs-2020` label to the PR so we can triage all the fixes later.

If you have questions, don't hesitate to message Ted Hudek on the Hopin platform during Writing Day.

Thank you for your help making our docs more accessible for everyone!

tcmetzger:
I'll tackle block 21!

AstroChoco:
I'm working on block 1.

AlexB-SheldonMFG:
I will take a shot at block 9.

caro-oviedo:
HI! I'll take block #28.

AlexB-SheldonMFG:
Finished block 9. I did not change designers\windows-forms-designer-overview file (line 200 in the files list). The procedure described in it seems to require exact knowledge the touchscreen UI to accurately describe. Cheers! -Alex.

technicaltidbits:
I'll take block 18

tcmetzger:
I'll get started on block 22.

tedhudek:
Thanks for everyone's time! We'll process these as soon as we can.

PR #5709, opened 2020-08-09
Update creating-a-windows-forms-toolbox-control.md

AlexB-SheldonMFG:
#5664, Write-the-Docs-2020

### DIFF
--- a/docs/extensibility/creating-a-windows-forms-toolbox-control.md
+++ b/docs/extensibility/creating-a-windows-forms-toolbox-control.md
@@ -29,7 +29,7 @@ The Windows Forms Toolbox Control template creates an undefined user control and
 
 1. Create a VSIX project named `MyWinFormsControl`. You can find the VSIX project template in the **New Project** dialog, by searching for "vsix".
 
-2. When the project opens, add a **Windows Forms Toolbox Control** item template named `Counter`. In the **Solution Explorer**, right-click the project node and select **Add** > **New Item**. In the **Add New Item** dialog, go to **Visual C#** > **Extensibility** and select **Windows Forms Toolbox Control**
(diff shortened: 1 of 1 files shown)


PRMerger17:
@AlexB-SheldonMFG : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Looks good; merging.

ghogen:
@ktoliver @AlexB-SheldonMFG Before merging this into the private repo, I checked and saw some problems.  There are places where "Select and hold" is used in the text, but it's not correct, the user has to double-click or double-tap.  In general, "select and hold" is the equivalent of right-click, not double-click.

ktoliver:
Hi @ghogen I'll be happy to make a PR to fix things. The guidance doesn't address double-click, but it's clearly an action that requires a mouse. Is the equivalent on a phone or tablet selecting and holding, or is it double-tapping?

ghogen:
@ktoliver Double tap is equivalent to double-lick - just tried it in Visual Studio for the scenario described in the article. Select and hold does something else - so for a button on a form, double-tap and double-click open the event handler in code.  Select and hold on the button brings up a menu.

ktoliver:
@ghogen Ah, OK. Sorry for my edits. I'll create a PR in the private repo to fix it (change "double-click"/"double-clicking" to "double-tap"/"double-tapping", correct?). Are you happy with replacing "right-click" with "select and hold (or right-click)" per the guidance given for the Write-the-Docs effort? (Looks right to me, FWIW.)

ghogen:
@ktoliver OK, tag me on the PR.  In some cases, we can use "open" or another verb instead of "double-click or double-tap" - that is more inclusive because keyboard actions also exist, usually selecting an item and hitting the Enter key, that are equivalent.

PR #5720, opened 2020-08-11
"double-click" fix update to PR 5709

ktoliver:
Fixes double-click/double-tap language

@ghogen - Do you want to change any of these instances to just "Open"? In the original PR, "double-click" was addressed, but revising to "open" is an acceptable solution.


ktoliver:
Thanks @ghogen

PR #5846, opened 2020-09-09
Docker Toolbox

philipwolfe:
Docker Toolbox is legacy and the link in this page redirects to Docker for Windows now.  I think it is confusing to list the legacy one first and so I think it should be removed.


PRMerger10:
@philipwolfe : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen (approved):

Issue #4258, opened 2019-10-31
ProjectReference metadata is incomplete


cdmihai:
There's more metadata that is not documented on this page, like metadata for controlling global properties and the output item type.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 97e663f7-3bac-61af-28bb-f0e471dcdca6
* Version Independent ID: 97985f56-e0ac-a03d-f651-b86cc2b85804
* Content: [Common MSBuild Project Items - Visual Studio](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2019#projectreference)
* Content Source: [docs/msbuild/common-msbuild-project-items.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/msbuild/common-msbuild-project-items.md)
* Product: **visual-studio-windows**
* Technology: **msbuild**
* GitHub Login: @Mikejo5000
* Microsoft Alias: **mikejo**

msebolt:
@cdmihai Thank you for the feedback, assigning this to the author for review.

rainersigwald:
Here's my half-written doc on this which is not ready to publish but may help for now: https://gist.github.com/rainersigwald/1fbd21a24a41b9b7c6204cb9cfcbb1cf

clisc:
Any update on this?  In the meantime, what is the best reference to use?

ghogen:
This was fixed in the referenced PR.

ghogen:
@rainersigwald 
If there's more to do here, please reactivate and let's get the remaining work identified.

PR #5848, opened 2020-09-09
Projectreference metadata

rainersigwald:
These have been long supported in common.targets but not documented here.

Not as complete as what I hoped to do for #4258, but a decent start.


PRMerger7:

ghogen (approved):

PR #5905, opened 2020-09-23
private beta for GH codespaces with VS

kendrahavens:
We are still in private beta for GitHub codespaces with Visual Studio. @gregvanl


PRMerger18:
@kendrahavens : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5968, opened 2020-10-12
Update MSBuild project SDK documentation

dsplaisted:
- Delete note about VS for Mac which is probably out of date (though it would be good to check)
- Refer to ".NET SDK" instead of ".NET CLI"


PRMerger18:
@dsplaisted : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen (approved):

PR #6087, opened 2020-11-11
Typo "Azure functions"→"Azure Functions"

hyoshioka0128:
https://docs.microsoft.com/en-us/visualstudio/mac/azure-functions?view=vsmac-2019

### DIFF
--- a/mac/azure-functions.md
+++ b/mac/azure-functions.md
@@ -11,7 +11,7 @@ ms.topic: how-to
 
 # Introduction to Azure Functions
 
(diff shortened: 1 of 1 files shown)


PRMerger19:
@hyoshioka0128 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@jmatthiesen - Could you review the proposed update to your article?

For reference, the internal cloud style guide says this:

"Use _Azure Functions_ as necessary to distinguish it as an Azure service. It's OK to use just _Functions_ when the context makes it clear you're referring to the Azure service.

To refer to something built using Functions or in a generic reference, use lowercase _function app_."

Thanks.

jmatthiesen (approved):
Good catch, thanks for the fix!

PR #6251, opened 2020-12-27
Outdated method

HSNHK:
hello
{% load static %} Django documents use this method.
{% load staticfiles %} This method is obsolete.
https://django.readthedocs.io/en/stable/howto/static-files/
https://docs.djangoproject.com/en/3.1/howto/static-files/

Unfortunately, this method does not work in newer versions of Django and causes errors.
The error is as follows:
TemplateSyntaxError at /
'staticfiles' is not a registered tag library


PRMerger9:
@HSNHK : Thanks for your contribution! The author(s) have been notified to review your proposed change.

JoshuaPartlow:
Hi @HSNHK - Thanks for submitting a PR. Right now the tutorials are tied to the templates and the templates are using an earlier version of Django. We'll take this into account thoough when we next refresh the Django content.

PR #6436, opened 2021-03-03
Update roslyn-version-support.md

jmarolf:
Add point release version info

### DIFF
--- a/docs/extensibility/roslyn-version-support.md
+++ b/docs/extensibility/roslyn-version-support.md
@@ -20,7 +20,16 @@ As an example, to ensure that your custom analyzer works on all versions of Visu
 
 | Roslyn package version | Minimum supported Visual Studio version |
 | - | - |
-| 3.x | Visual Studio 2019 |
+| 3.9.0 | Visual Studio 2019 16.9 |
+| 3.8.0 | Visual Studio 2019 16.8 |
+| 3.7.0 | Visual Studio 2019 16.7 |
+| 3.6.0 | Visual Studio 2019 16.6 |
+| 3.5.0 | Visual Studio 2019 16.5 |
+| 3.4.0 | Visual Studio 2019 16.4 |
+| 3.3.1 | Visual Studio 2019 16.3 |
+| 3.2.1 | Visual Studio 2019 16.2 |
+| 3.1.0 | Visual Studio 2019 16.1 |
+| 3.0.0 | Visual Studio 2019 RTM |
 | 2.10.0 | Visual Studio 2017 version 15.9 |
 | 2.9.0 | Visual Studio 2017 version 15.8 |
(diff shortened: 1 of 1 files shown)


PRMerger16:
@jmarolf : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/extensibility/roslyn-version-support.md
```diff
@@ -20,7 +20,16 @@ As an example, to ensure that your custom analyzer works on all versions of Visu
 
 | Roslyn package version | Minimum supported Visual Studio version |
```
Youssef1313, line 21: @jmarolf Can we have a third column for Minimum supported SDK version in the case of command-line builds?

jmarolf: .NET SDK versions do not line up in any real way with VS versions. We would need many more columns to explain this.

in SDK version 2.1.500, `2.1` referes to the maximum runtime version that this sdk can build against and `500` means the fifth release of the SDK.

Youssef1313: Isn't the minimum SDK version the same one ships with the specific Visual Studio version?
For example for Roslyn packages 3.9.x it's Visual Studio 16.9 and .NET SDK 5.0.200?

jmarolf: hmm, so if you got to dot.net and download the .NET 5 SDK it will be 5.0.103 which is compatible with 3.8.3 compilers and newer. If you use the SDK that is installed by VS 2019 16.9 you get 5.0.200 which is compatible with 3.9.0. So If you just want to know what SDK ships with which Visual Studio that's an easy mapping. If you want to know which compiler to target so users of a particular SDK (say .NET 5 SDKs)  that becomes a more complex question that frankly needs a different table altogether.

Youssef1313: > If you want to know which compiler to target so users of a particular SDK (say .NET 5 SDKs) that becomes a more complex question that frankly needs a different table altogether.

Yeah this is what I meant. But I thought it's the same as "see the roslyn version supported by whatever VS version that shipped this particular SDK".

Now I'm more confused to be honest.

To be more clear, the intent here is when authoring analyzers. When I upgrade roslyn version in my analyzers:

- What VS version will be required for analyzer consumers.
- What SDK version will be required for analyzer consumers in command-line builds.

jmarolf: I get the problem we need to solve here (and we do need something). We may want a different doc page for analyzer authors as things get complicated.

jmarolf:
ping @acangialosi and @leslierichardson95

canton7:
As far as I know, this is the only place on the internet where this information is documented, and it's in a PR which was closed for no apparent reason.

Can we please open this up and get it merged?

PR #6468, opened 2021-03-10
Update index.yml

IEvangelist:
This logo is inaccurate - I think we could use the official logo SVG which I believe is [available here](https://docs.microsoft.com/en-us/media/logos/logo_octokitty.svg).

image

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -52,7 +52,7 @@ productDirectory:
         - url: https://code.visualstudio.com
           text: Product page
     - title: GitHub Codespaces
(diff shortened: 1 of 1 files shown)


PRMerger18:
@IEvangelist : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Hi @IEvangelist - This pull request was opened in the public repo. Would you instead submit it in the private repo, so the proposed image can be viewed on staging? You can at-mention @ghogen, if you'd like, before signing off in your private repo PR. Thanks! We'll close this PR.

Issue #6748, opened 2021-06-08
Installing BuildTools creates a separate VS installation rather than updating the existing installation


mthalman:
The Dockerfile (shown below) provided in the documentation explicitly provides an install path separate from the already existing installation of BuildTools in the base image `mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019`. This is wasteful because it creates a separate installation with duplicate files that could have been reused if the existing installation had just been updated with the desired workload.

```Dockerfile
# escape=`

# Use the latest Windows Server Core image with .NET Framework 4.8.
FROM mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019

# Restore the default Windows shell for correct batch processing.
SHELL ["cmd", "/S", "/C"]

# Download the Build Tools bootstrapper.
ADD https://aka.ms/vs/15/release/vs_buildtools.exe C:\TEMP\vs_buildtools.exe

# Install Build Tools with the Microsoft.VisualStudio.Workload.AzureBuildTools workload, excluding workloads and components with known issues.
RUN start /wait C:\TEMP\vs_buildtools.exe --quiet --wait --norestart --nocache `
    --installPath C:\BuildTools `
    --add Microsoft.VisualStudio.Workload.AzureBuildTools `
    --remove Microsoft.VisualStudio.Component.Windows10SDK.10240 `
    --remove Microsoft.VisualStudio.Component.Windows10SDK.10586 `
    --remove Microsoft.VisualStudio.Component.Windows10SDK.14393 `
    --remove Microsoft.VisualStudio.Component.Windows81SDK `
 || IF "%ERRORLEVEL%"=="3010" EXIT 0

# Define the entry point for the Docker container.
# This entry point starts the developer command prompt and launches the PowerShell shell.
ENTRYPOINT ["C:\\BuildTools\\Common7\\Tools\\VsDevCmd.bat", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
```

You can see that separate VS installations are in the container by running the installer's VSWhere tool:

(I've pared down the JSON results for readability)
```console
C:\>"C:\Program Files (x86)\Microsoft Visual Studio\installer\vswhere" -products * -all -format json
[
  {
    "instanceId": "13d925de",
    "installationPath": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools"
  },
  {
[...]

PR #6754, opened 2021-06-09
Update how VS Build Tools is installed

mthalman:
In the VS 2019 example Dockerfile, it creates a separate installation of VS Build Tools rather than modify the existing installation of Build Tools from the sdk base image.  I've updated it to make use of the existing installation instead.

I've also modified the way in which the bootstrapper is downloaded.  The previous way using the `ADD` instruction caused unnecessary bloat in the resulting image because the vs_buildtools.exe is contained in an image layer (even deleting the file in a separate layer wouldn't solve this because the layer in which it was downloaded still exists).  I've modified the pattern to download the bootstrapper with curl and delete the bootstrapper after installation, all in the same layer.  This prevents the bootstrapper exe existing within the built image.  I applied this same pattern to the VS 2017 Dockerfile.

Fixes https://github.com/MicrosoftDocs/visualstudio-docs/issues/6748

### DIFF
--- a/docs/install/build-tools-container.md
+++ b/docs/install/build-tools-container.md
@@ -64,18 +64,22 @@ Save the following example Dockerfile to a new file on your disk. If the file is
    # Restore the default Windows shell for correct batch processing.
    SHELL ["cmd", "/S", "/C"]
 
-   # Download the Build Tools bootstrapper.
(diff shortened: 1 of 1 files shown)


PRMerger18:
@mthalman : Thanks for your contribution! The author(s) have been notified to review your proposed change.

j-martens:
#sign-off

PR #6830, opened 2021-07-05
Clarify performance impacts of "Simplify LINQ expression" refactoring

### LINKED ISSUE #6769: Simplify LINQ expression: Example is actually reducing performance
The documentation claims:

> **Why:** Removing the unnecessary call to the Enumerable for the .Where() method improves performance and readability.

This is actually not correct in all cases but especially not in the case of the given example, see https://github.com/dotnet/runtime/issues/19382#issuecomment-467612364.

As pointed out in https://github.com/dotnet/runtime/issues/19382#issuecomment-734524103 this will not get fixed in .NET because of the very high complexity and the availability of a workaround. This workaround is being marked as slower in this Refactoring even though it is faster.

As there are still cases where replacing `.Where(pred).First()` with `.First(pred)` can actually be faster (mainly custom `IQueryable<T>` implementations) I don't necessarily mean that the Refactoring needs to be removed but the example is not aligning with the performance claim. As it is also not reasonably possible to check if the `IEnumerable<>` is a `List<T>` or `T[]`, I would suggest to remove the performance claim altogether and add a remark that `List<T>` and `T[]` will be negatively impacted by this change.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 775e2113-11a7-71e7-6f22-10877334492f
* Version Independent ID: 69e646aa-22a0-487a-be05-415b7b5f1774
* Content: [Simplify LINQ expression - Visual Studio](https://docs.microsoft.com/en-us/visualstudio/ide/reference/simplify-linq-expression?view=vs-2019&preserve-view=true)
* Content Source: [docs/ide/reference/simplify-linq-expression.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/ide/reference/simplify-linq-expression.md)
* Product: **visual-studio-windows**
* Technology: **vs-ide-general**
* GitHub Login: @m-redding
* Microsoft Alias: **midumont**

mhartkorn:
Fixes #6769

### DIFF
--- a/docs/ide/reference/simplify-linq-expression.md
+++ b/docs/ide/reference/simplify-linq-expression.md
@@ -1,7 +1,7 @@
 ---
 title: Simplify LINQ expression
 description: "This refactoring is used to remove unnecessary calls to the Enumerable for the Where method."
-ms.date: 08/12/2020
+ms.date: 07/05/2021
 ms.topic: reference
 author: m-redding
 ms.author: midumont
@@ -21,7 +21,7 @@ This refactoring applies to:
 
 **When:**  All instances where the method calls `Single()`, `SingleOrDefault()`, and so on, doesn't have any arguments and is preceded by a `Where()` expression. The input to the `Where()` expression cannot be constructed as an expression tree.
 
-**Why:** Removing the unnecessary call to the Enumerable for the `.Where()` method improves performance and readability.
+**Why:** Removing the unnecessary call to the Enumerable for the `.Where()` method improves readability and in some cases performance, see remarks.
 
 ## How-to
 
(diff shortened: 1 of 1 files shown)


TerryGLee:
@mikadumont FYI

j-martens:
#sign-off

PRMerger7:
@j-martens: I'm sorry - only the author of this article, @m-redding, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #6976, opened 2021-08-17
Update that VS2022 will eventually have some bare-bones support for CUIT and Web Load Test

kendrahavens:


kendrahavens:
Moving to pr repo

PR #7043, opened 2021-09-13
error on nuget reinstall command

Thieum:
It's a single dash, instead of a double dash, see reference in article.


PRMerger8:
@Thieum : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7306, opened 2021-11-17
Update general-debugging-options-dialog-box.md

hharshada:
Removing the setting options which are not applicable with VS2022

### DIFF
--- a/docs/debugger/general-debugging-options-dialog-box.md
+++ b/docs/debugger/general-debugging-options-dialog-box.md
@@ -139,16 +139,6 @@ Controls the direction in which stacks are displayed in the **Parallel Stacks** 
 **Ignore GPU memory access exceptions if the data written didn't change the value**:
 Ignores race conditions that were detected during debugging if the data didn't change. For more information, see [Debugging GPU Code](../debugger/debugging-gpu-code.md).
 
-**Use Managed Compatibility Mode**:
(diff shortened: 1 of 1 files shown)


PRMerger17:
@hharshada : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Mikejo5000 (approved):

Mikejo5000 (changes requested):
@hharshada Actually, I rechecked this. Can you restore the old content and put into a version-monikered section? It would look like this:
::: moniker range="<=vs-2022"
**Use Managed Compatibility Mode**
 ...old content
::: moniker-end
I'm assuming this option was present in VS 2019.

hharshada:
Closing this as new PR is created

Mikejo5000:
@hharshada Sure, that's fine. You might also just copy the markdown from the Files Changed tab above (it shows the old and new markdown) back into the PR. Here's the markdown:

**Use Managed Compatibility Mode**:
Replaces the default debugging engine with a legacy version to enable these scenarios:

- You are using a .NET language other than C#, Visual Basic, or F# that provides its own Expression Evaluator (this includes C++/CLI).

- You want to enable Edit and Continue for C++ projects during mixed mode debugging.

> [!NOTE]
> Choosing Managed Compatibility mode disables some features that are implemented only in the default debugging engine. The legacy debugging engine was replaced in Visual Studio 2012.

PR #7689, opened 2022-02-18
Use the correct path to the .targets file for v17

mrlacey:
Just a quick simple change so the right path is shown when viewing this document for use with VS2022


PRMerger10:
@mrlacey : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Issue #6843, opened 2021-07-09
Document makes incorrect statement regarding the performance of an application that uses the SuppressMessage attribute


StingyJack:
From the document...

> You should not use in-source suppressions on release builds, to prevent shipping the in-source suppression metadata accidentally. Additionally, because of the processing cost of in-source suppression, the performance of your application can be degraded.

The suppress message attributes are not included or represented in the IL of the compiled class file in either Debug or Release mode. How can something that is not even present negatively impact runtime application performance?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 517e8321-aa13-0814-3398-9606d7c2664b
* Version Independent ID: 59efa3c4-ceaf-a384-94af-0f76584efe3f
* Content: [Suppress code analysis violations - Visual Studio (Windows)](https://docs.microsoft.com/en-us/visualstudio/code-quality/in-source-suppression-overview?view=vs-2019#in-source-suppression-and-the-suppressmessage-attribute)
* Content Source: [docs/code-quality/in-source-suppression-overview.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/code-quality/in-source-suppression-overview.md)
* Product: **visual-studio-windows**
* Technology: **vs-ide-code-analysis**
* GitHub Login: @mikadumont
* Microsoft Alias: **midumont**

StingyJack:
If the author has not changed the errant content, why would this be considered "stale"?

j-martens:
Thanks for the fix, @StingyJack.  We will now proceed to close the issue.

PR #7792, opened 2022-03-17
Removing stmt that cant be correct as is #6843

StingyJack:
An attribute that is not present in the compiled output cannot negatively affect the performance of a program


PRMerger-2:
@StingyJack : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7913, opened 2022-04-15
Add documentation for new task authoring guidance

baronfel:
As part of https://github.com/dotnet/msbuild/pull/7520, we have developed new guidance for MSBuild Task authors to make their tasks (and MSBuild overall) more resilient in the face of widely-varying task dependencies. We're using a mechanism developed by the dotnet/runtime team to ensure assembly load isolation, and in preparation we need to have guidance for task authors about what we think the best way to package is, and how to do so with their tasks.

This is that documentation!  It's been put in to practice with a tasks package that I maintain (https://github.com/ionide/KeepAChangelog) in conjunction with members of the MSBuild engineering team, so we're pretty happy with it.

I would love editorial review of this work, as I'm sure I have not kept to the same 'voice' or tone throughout.  I also made a stylistic change to the XML segments of this documentation - rather than copy the entire project file (since it would get long quite quickly), I have used ellipsis to elide sections that haven't changed.  Let me know what you think!

### DIFF
--- a/docs/msbuild/tutorial-custom-task-code-generation.md
+++ b/docs/msbuild/tutorial-custom-task-code-generation.md
@@ -56,50 +56,50 @@ The first step is to create the MSBuild custom task. Information about how to [w
 
 1. Add three properties. These properties define the parameters of the task that users set when they use the task in a client project:
 
-   ```csharp
-        //The name of the class which is going to be generated
-        [Required]
-        public string SettingClassName { get; set; }
+	```csharp
+		  //The name of the class which is going to be generated
+		  [Required]
+		  public string SettingClassName { get; set; }
 
-        //The name of the namespace where the class is going to be generated
-        [Required]
-        public string SettingNamespaceName { get; set; }
+		  //The name of the namespace where the class is going to be generated
+		  [Required]
+		  public string SettingNamespaceName { get; set; }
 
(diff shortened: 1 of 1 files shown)


PRMerger-2:
@baronfel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

baronfel:
@rainersigwald @Forgind would you both mind checking this for technical accuracy?

PRMerger18:
@baronfel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jborsecnik:
@ghogen, if you approve these changes, they need to be moved to the private repo for staging, since they involve potential format changes. Can you move them?

ghogen:
@jborsecnik 
OK, I'll move the PR into the private repo and there I can review and suggest edits to the work,

Forgind (commented):
I think the only big comment I have is that we want people to include .deps.json files, especially if they have (complicated) dependencies to resolve, but if they don't have any dependencies (or we can assume they're loaded for some reason, etc.) they aren't necessary.

PRMerger17:
@baronfel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen:
Completed in private repo - see link above.

PR #7960, opened 2022-04-30
Fix FxCop.Usage.UseParams code example

xparadoxical:
Due to incorrect condition (or iterator), this `for` loop iterates over at most half of the argument list, and also `i` is unnecessary for what the body does. I replaced it with a `while` loop.


Court72:
@ghogen

Can you review the proposed changes? *IMPORTANT*: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

ghogen (approved):

PR #8102, opened 2022-06-02
MSBuild: don't use both Before and AfterTargets

rainersigwald:
Clarify the interaction between BeforeTargets and AfterTargets and don't encourage using them together, which rarely does what the author expected.

@ghogen I'm adding you to the internal mail thread on this for context.

### DIFF
--- a/docs/msbuild/target-build-order.md
+++ b/docs/msbuild/target-build-order.md
@@ -112,7 +112,7 @@ Alternately, specify the order as
 </Target>
 ```
 
-It is not generally useful to specify _both_ `BeforeTargets` and `AfterTargets` on the same target. As described in the next section, only the first target encountered will cause the new target to run.
+It is not generally useful to specify *both* `BeforeTargets` and `AfterTargets` on the same target. As described in the next section, only the first target encountered will cause the new target to run.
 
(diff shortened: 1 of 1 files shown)


PRMerger-2:

docs/msbuild/target-build-order.md
```diff
@@ -99,12 +99,21 @@ Consider the following script.
...
 
+Alternately, specify the order as
+
+```xml
+<Target Name="Optimize" BeforeTargets="Link">
+    <Message Text="Optimizing" />
+</Target>
+```
+
+It is not generally useful to specify _both_ `BeforeTargets` and `AfterTargets` on the same target. As described in the next section, only the first target encountered will cause the new target to run.
```
ghogen, line 115:
```suggestion
It is not generally useful to specify *both* `BeforeTargets` and `AfterTargets` on the same target. As described in the next section, only the first target encountered will cause the new target to run.
```

PRMerger-2:
@ghogen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ghogen:
#sign-off

ghogen (approved):

PR #8134, opened 2022-06-09
Indenting link descriptions

jayallen:
Without proper indentation, each lined seemed like to be describing the following link and not the preceding one.  Adding two spaces of whitespace at the end of the bulleted line (which inserts a newline instead of soft wrapping) and four spaces in front of the description (for the indentation) fixes this issue.


PRMerger13:
@jayallen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #8373, opened 2022-08-17
Update Roslyn's Visual Studio version mapping table.

JoeRobich:
Updates Roslyn version table with specific package numbers for VS2019 and VS2022.


JoeRobich:
CC: @ghogen

PR #8397, opened 2022-08-26
Fix a small typo in `MSBuild reserved and well-known properties`

just-ero:


Jak-MS:
@just-ero - thanks for the edit!

PR #8780, opened 2022-12-20
Update windows-forms-designer-overview.md

simon6502:
Extra erroneous word removed.
BEFORE: When using sing **Shift**+click
AFTER: When using **Shift**+click


Court72:
Looks good; thanks, @simon6502

PR #8840, opened 2023-01-12
Adding version specifier to paragraph on filters

webreidi:
Didn't realize the same page was used for different versions. Need to remove the filters from VS2019


Jak-MS:
@Mikejo5000 
- Can you review this PR?

NOTE: Because this adds monikers we'll need to check this on staging before it can be merged, and to do that the commits will need to be moved to the private repo. Can you do that? (let us know if you'd like assistance.)


#label:"aq-pr-triaged"

Mikejo5000 (approved):

Mikejo5000:
#sign-off @webreidi Thanks for the correction!

Jak-MS:
@Mikejo5000 @webreidi - usually we don't allow prs that add/change/remove monikers without checking staging first, and we only have staging in the private repos. that said, the syntax looks correct and I think the chance of this rendering incorrectly is low. I'm going to merge it... with the request that @webreidi check this on the live site after it's updated to make sure it looks right. Thanks.

PR #8844, opened 2023-01-13
Add note that the Just-In-Time component is required

AWilco:
The Just-In-Time debugger is only installed if you install certain feature packages for Visual Studio (E.G. C# development or C++ Desktop development). If it is not installed the Just-In-Time section is missing from Options > Debug. It would be helpful to note that if this doesnt exist the user should ensure the component is installed. (Or potentially update VS to explain that there are no options because it is not installed).

I was trying to debug an IIS website crash, and so had only installed the ASP.NET development module. Without clicking on another feature package in the installer and seeing that there is a Just-In-Time debugger option that is added it is not obvious that this is a component that needs to be installed.


Court72:
@Mikejo5000 

When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

Mikejo5000:
#sign-off Thanks for the doc update, @AWilco !

PR #8861, opened 2023-01-17
Update Django tutorial with re_path call

MarissaB:
https://github.com/MicrosoftDocs/visualstudio-docs/issues/8847


cwebster-99 (approved):
Thank you for the bug report and contribution!

PR #8987, opened 2023-02-23
Mention Private metadata for ProjectReference

JanKrivanek:
## Context

`ProjectReference` item accepts `Private` metadata with the same behavior and meaning as `Reference` item.
The later has the metadata documented the former has not - so closing the gap.


ghogen (approved):

ghogen:
#sign-off

PR #9086, opened 2023-03-26
Add note for fixing connection string error

swizkon:
Without quotes the connection parameter isnt parsed correctly


swizkon (commented):
#sign-off Good to go

swizkon:
#sign-off Good to go and tested with GH actions and Postgres on Azure

ghogen:
This is a good point, we are working on a fix and will publish that shortly! Thanks!

PR #9385, opened 2023-06-30
The Exec task calls /bin/sh on Linux

gqqnbig:
I was wondering how the Exec task works on Linux so I read [the source code](https://github.com/dotnet/msbuild/blob/main/src/Tasks/Exec.cs) and found it calls `/bin/sh` (rather than, say `bash`) with the specified command.

Therefore I think it's worth noting that the Exec task calls /bin/sh on Linux on the documentation.

### DIFF
--- a/docs/msbuild/exec-task.md
+++ b/docs/msbuild/exec-task.md
@@ -51,7 +51,7 @@ The following table describes the parameters for the `Exec` task.
 
 This task is useful when a specific MSBuild task for the job that you want to perform is not available. However, the `Exec` task, unlike a more specific task, cannot do additional processing or conditional operations based on the result of the tool or command that it runs.
 
-The `Exec` task calls *cmd.exe* instead of directly invoking a process.
+Instead of directly invoking a process, the `Exec` task calls *cmd.exe* or */bin/sh* depending on the OS.
 
 The parameters `IgnoreExitCode` and `IgnoreStandardErrorWarningFormat` affect the conditions under which the task returns `false`, indicating an error. With the default settings (`false` for both), the `Exec` task indicates a failure (returns `false`) either if the executable has a non-zero exit code, or if a diagnostic message is found in the executable's standard error stream. If you only want `Exec` to indicate failure if the executable returns a non-zero exit code, then set `IgnoreStandardErrorWarningFormat` to `true`.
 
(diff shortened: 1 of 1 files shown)


docs/msbuild/exec-task.md
```diff
@@ -51,7 +51,7 @@ The following table describes the parameters for the `Exec` task.
 
 This task is useful when a specific MSBuild task for the job that you want to perform is not available. However, the `Exec` task, unlike a more specific task, cannot do additional processing or conditional operations based on the result of the tool or command that it runs.
 
-The `Exec` task calls *cmd.exe* instead of directly invoking a process.
+Instead of directly invoking a process, the `Exec` task calls *cmd.exe* or */bin/sh* depending on the OS.
```
ghogen, line 54:
```suggestion
Instead of directly invoking a process, the `Exec` task calls *cmd.exe* or */bin/sh*, depending on the operating system.
```

docs/msbuild/exec-task.md
```diff
@@ -51,7 +51,7 @@ The following table describes the parameters for the `Exec` task.
 
 This task is useful when a specific MSBuild task for the job that you want to perform is not available. However, the `Exec` task, unlike a more specific task, cannot do additional processing or conditional operations based on the result of the tool or command that it runs.
 
-The `Exec` task calls *cmd.exe* instead of directly invoking a process.
+Instead of directly invoking a process, the `Exec` task calls *cmd.exe* or */bin/sh*, depending on the operating system.
```
ghogen, line 54:
```suggestion
Instead of directly invoking a process, the `Exec` task calls *cmd.exe*  on Windows, or *sh* otherwise.
```

ghogen: It calls "sh" on non-Windows operating systems, but I think this would resolve to whatever sh is installed when the path is searched. It could be \bin\sh but if sh is installed somewhere else and that precedes \bin on the path, then it would be used.

ghogen: See https://github.com/dotnet/msbuild/blob/467f900dd0ae91d2ef5d3ab523cb11dbd148f320/src/Tasks/Exec.cs#L531

ghogen: @rainersigwald. We're trying to update the wording regarding the shell that is invoked by the Exec task on a non-Windows operating system.

rainersigwald: Your analysis seems accurate though I wonder if we _should_ call fully-qualified `/bin/sh` instead of `sh` from PATH. But I like the update with your suggestion.

rainersigwald (approved):

ghogen:
#sign-off

PR #9859, opened 2023-12-05
fix: Installer command line, Anchor to use channelUri

Styxxy:
The anchor on the page is all lower case, in certain browsers the Anchoring seems to be case-sensitive.


ShannonLeavitt:
Looks good. Thanks @Styxxy

PR #9898, opened 2023-12-15
Description in wrong column for get[Item|Property|TargetResult]

jrdodds:
Update msbuild-command-line-reference.md to change -getItem, -getProperty, -getTargetResult to move description to the correct column.


Court72:
@jrdodds, Thanks for the formatting fix!

PR #9932, opened 2024-01-05
Update build-tools-container.md to default to 64-bit builds

ofek:
As a novice when it comes to building on Windows, it took a significant amount of research to find out why I was getting missing symbols when compiling basic extension modules for Python.

https://learn.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line#remarks

> When you install a C++ workload in the Visual Studio installer, it always installs 32-bit, x86-hosted, native and cross compiler tools to build x86 and x64 code.

I couldn't find out the proper flag for 64-bit so I used the dedicated batch file.

### DIFF
--- a/docs/install/build-tools-container.md
+++ b/docs/install/build-tools-container.md
@@ -81,7 +81,7 @@ Save the following example Dockerfile to a new file on your disk. If the file is
 
    # Define the entry point for the docker container.
    # This entry point starts the developer command prompt and launches the PowerShell shell.
-   ENTRYPOINT ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\Common7\\Tools\\VsDevCmd.bat", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
+   ENTRYPOINT ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Auxiliary\\Build\\vcvars64.bat", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
    ```
 
    > [!TIP]


Court72:
@anandmeg

When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

anandmeg:
@ofek, thanks for this feedback. We're consulting with the product team and will update the docs per their input. Thanks again!

ofek:
Can you please explain a little bit further? Alternatively, it would perhaps be better to keep this open and then close when the change happens otherwise it appears externally that the feedback is simply ignored.

anandmeg:
@ofek, per input from the product team, vcvars64.bat is the old version of the developer command prompt.  Also, VSVars64.bat file might not be there if the user is only building .NET stuff.  For VsDevCmd to target 64-bit, you can use the flag
-arch=amd64. We'll revise and update the docs accordingly. 
Thanks!

docs/install/build-tools-container.md
```diff
@@ -81,7 +81,7 @@ Save the following example Dockerfile to a new file on your disk. If the file is
 
    # Define the entry point for the docker container.
    # This entry point starts the developer command prompt and launches the PowerShell shell.
-   ENTRYPOINT ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\Common7\\Tools\\VsDevCmd.bat", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
+   ENTRYPOINT ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Auxiliary\\Build\\vcvars64.bat", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
```
ofek, line 84:
```suggestion
   ENTRYPOINT ["C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Auxiliary\\Build\\vcvarsall.bat", "x64", "&&", "powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]
```

ofek:
I updated, thanks! Does this look like how you imagined now?

anandmeg:
@ofek, thanks for your contribution & feedback. We've updated the docs per product team's input. The changes should be live later today. Thanks!

PR #10101, opened 2024-03-06
Disambiguate from-scratch project from csproj

rainersigwald:
MSBuild doesn't usually have very strong coupling to file extension, but some things (parts of the .NET SDK, parts of Visual Studio, and lots of third-party code) will assume that a file named `*.csproj` will walk and quack like a "real"/"standard" `.csproj`.

Since that's not the case for the built-from-scratch-as-an-exercise project in this walkthrough, let's name it something else.

This tripped up someone who was following the walkthrough intending to build a new not-C# project for actual production use: https://mastodon.social/@mcc/112050557130975505

### DIFF
--- a/docs/msbuild/walkthrough-creating-an-msbuild-project-file-from-scratch.md
+++ b/docs/msbuild/walkthrough-creating-an-msbuild-project-file-from-scratch.md
@@ -16,6 +16,8 @@ Programming languages that target the .NET Framework use MSBuild project files t
 
 > [!NOTE]
 > This article is appropriate if you want to learn the basic fundamentals of how MSBuild works independently of any SDK. Building with an SDK, such as when you use `dotnet build` or you add the `Sdk` attribute to the root project element, is not covered in this article. See [.NET Project SDKs](/dotnet/core/project-sdk/overview).
+>
+> The build logic imported by standard `.csproj` files supports many more options and a much more complex build process than this example.
 
  For information about creating a project file for a C++ project, see [MSBuild (C++)](/cpp/build/msbuild-visual-cpp).
 
@@ -99,7 +101,7 @@ If you're running on Windows, you can use the **Developer Command Prompt for Vis
 
 ### To create a minimal MSBuild project file
 
-1. In the text editor, create a new file *HelloWorld.csproj* and enter the following code:
+1. In the text editor, create a new file *HelloWorld.fromscratchproj* and enter the following code:
 
     ```xml
     <Project>
@@ -124,7 +126,7 @@ If you're running on Windows, you can use the **Developer Command Prompt for Vis
     <Csc Sources="@(Compile)"/>
     ```
 
-1. Save this project file and name it *Helloworld.csproj*.
+1. Save this project file and name it *Helloworld.fromscratchproj*.
 
 Your minimal project file should resemble the following code:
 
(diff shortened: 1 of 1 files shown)


docs/msbuild/walkthrough-creating-an-msbuild-project-file-from-scratch.md
```diff
@@ -99,7 +101,7 @@ If you're running on Windows, you can use the **Developer Command Prompt for Vis
 
 ### To create a minimal MSBuild project file
 
-1. In the text editor, create a new file *HelloWorld.csproj* and enter the following code:
+1. In the text editor, create a new file *HelloWorld.fromscratchproj* and enter the following code:
```
rainersigwald, line 104: @ghogen do you think we should explicitly call out that both filename and extension are arbitrary, though ending in `proj` is conventional? Not sure if that's too much to explain in this context.

Jak-MS:
@ghogen 
- Can you review this PR?


#label:"aq-pr-triaged"

dionrhys:
As a data point: at Stack Overflow we use ".msbuild" as the file extension for generic MSBuild project files, e.g.

- `tc.stackoverflow.public.msbuild`
- `tc.stackoverflow.enterprise.msbuild`
- `tc.apiv3.enterprise.msbuild`
- etc.

When I joined Stack, I found the convention quite intuitive because "oh it ends in .msbuild, that must mean it's an input to MSBuild".

If I look at a ".proj" or ".fromscratchproj" file, it wouldn't be immediately obvious what this file is for (counter to how ".csproj" and ".vcxproj" _would_ be for example) - it could be a project file for any kind of build system without looking inside the file.

---

_**However,** also,_ if I search for ".proj file" online, the first result I get is <https://learn.microsoft.com/en-us/aspnet/web-forms/overview/deployment/web-deployment-in-the-enterprise/understanding-the-project-file>, which explains:

> # Project File Naming Conventions
>
> When you create your own project files, you can use any file extension you like. However, to make your solutions easier for others to understand, you should use these common conventions:
>
> - Use the .proj extension when you create a project file that builds projects.
> - Use the .targets extension when you create a reusable project file to import into other project files. Files with a .targets extension typically don't build anything themselves, they simply contain instructions that you can import into your .proj files.

That documentation is under _ASP.NET Core -> Web Forms -> Guidance -> Deployment -> ASP.NET 4 - Enterprise Deployment Series 1 -> Understanding the Project File_, which is a deep and old-looking reference admittedly, but being an extension already in use in documentation and has good search results could be useful for folks searching for this information in future.

---

I think giving an example of `.fromscratchproj` in the documentation could come across cumbersome for folks trying to create these kind of MSBuild "meta" projects for their solutions. I think documenting a shorter name like ".msbuild" or ".proj" could make the topic more approachable / intuitive.

rainersigwald:
Yeah, just `.proj` might be a good option.

`.msbuild` makes sense, but the Microsoft convention is generally:

* Projects that can be built end in `*proj`
* A project that "doesn't do much" is just `.proj` (e.g. `dirs.proj` for traversal projects that only call the MSBuild task on other projects).
* A project class should get its own extension (`.csproj` for C#, `.esproj` for JavaScript, `.vcxproj` for C++)

IIRC parts of VS respect `.msbuildproj`. I'm not aware of anything that has special behavior for `.msbuild` (even though in retrospect that is a good choice).

docs/msbuild/walkthrough-creating-an-msbuild-project-file-from-scratch.md
ghogen: @rainersigwald 
I think what you say should be mentioned in the top-level MSBuild overview, rather than here. I think for our purposes here, we could use the extension ".proj" or as you did, ".fromscratchproj".

ghogen: Internal PR created. https://github.com/MicrosoftDocs/visualstudio-docs-pr/pull/12284

ghogen (approved):

PR #10116, opened 2024-03-12
Make it clear that you cannot filter the list of tests that get displayed

rfrancioso-gpsw:
- In my pipeline I was trying to output a clean list of all the tests that would get run, before actually running them.
- When I do run tests, I use a filter, so I was also trying to use the same filter when listing tests.
- I spent over an hour trying to understand why the filter I was specifying was not working.
- In the end it ended up being because a specified filter has no effect when listing tests; only when running tests.

I've updated the documentation to make this clear.  It would be nice if VSTest was also updated at some point in the future to actually support filtering when listing.

### DIFF
--- a/docs/test/vstest-console-options.md
+++ b/docs/test/vstest-console-options.md
@@ -41,7 +41,7 @@ The following table lists all the options for *VSTest.Console.exe* and short des
 |**/TestCaseFilter:[*expression*]**|Run tests that match the given expression.<br /><Expression\> is of the format <property\>=<value\>[\|<Expression\>].<br />Example: `/TestCaseFilter:"Priority=1"`<br />Example: `/TestCaseFilter:"TestCategory=Nightly|FullyQualifiedName=Namespace.ClassName.MethodName"`<br />The **/TestCaseFilter** command-line option cannot be used with the **/Tests** command-line option. <br />For information about creating and using expressions, see [TestCase filter](https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md).|
 |**/?**|Displays usage information.|
 |**/Logger:[*uri/friendlyname*]**|Specify a logger for test results. Specify the parameter multiple times to enable multiple loggers.<br />Example: To log results into a Visual Studio Test Results File (TRX), use<br />**/Logger:trx**<br />**[;LogFileName=\<Defaults to unique file name>]**|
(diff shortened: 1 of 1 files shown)


Jak-MS:
@Mikejo5000 
- Can you review this PR?


#label:"aq-pr-triaged"

Mikejo5000:
@Evangelink Does anyone on the team need to review this PR? It looks OK to me.

Evangelink:
I would rather update the `/TestCaseFilter` entry to mention it's not working with list tests but that's a personal feeling so I am not blocking here.

Mikejo5000 (approved):

Mikejo5000:
#sign-off OK, thanks. It can be updated again later.

rfrancioso-gpsw:
@Evangelink Related to test case filtering...

Would anyone on your team be able to answer this by any chance?  I'm really hoping there's a way to accomplish this, but it doesn't seem like there is.  If there isn't it would be nice if this functionality could be added in a future release.

https://stackoverflow.com/questions/78144390/how-to-run-all-unit-tests-that-dont-have-a-trait

Evangelink:
Sure thing, it will depends on the test framework you are using so I'll let you have a look at this doc page https://learn.microsoft.com/dotnet/core/testing/selective-unit-tests

rfrancioso-gpsw:
Thanks for the link @Evangelink.  I took a look, but all the operators that are called out there don't seem to handle the following scenario.  Imagine you have two categories of tests:
- Regular Unit Tests
    -  These **don't** have an assigned category (because there's nothing special about them).
- Slow Tests
    - These **do** have an assigned category: (`[TestCategory("Slow")]`)

For a PR build, you now want to run all Unit Tests that **don't** have an assigned category.  These don't work:
`/TestCaseFilter:"TestCategory="`
`/TestCaseFilter:"TestCategory=''"`

I know you can do this instead: `/TestCaseFilter:"TestCategory!=Slow"`

But imagine that in addition to there being "Slow", there's also 10 other categories that need to be excluded as well.  And when new categories get added in the future, you don't want to have to keep updating your filter.

Is there no way to simply select all tests that **haven't** been assigned a category?

Evangelink:
Sadly it's not possible and it's a limitation of VSTest. I'll try to find the ticket mentioning exactly this issue and link it here

PR #10137, opened 2024-03-19
Update bridge-to-kubernetes-vs.md

ProfessorMc:
https://github.com/Azure/Bridge-To-Kubernetes/pull/621 removed the todo app an moved it to a custom fork for dependabot.  Updated the link to the fork.


ghogen:
@ProfessorMc @qpetraroia 
I don't think we should referencing any private forks from the documentation. Can you explain the reasoning here, also I've added a Bridge product team member for review.

PR #10162, opened 2024-03-30
Fix code example not appearing under the bullet point

adegeo:
Indented the snippet to align to the bullet point.


ShannonLeavitt:
Looks good. Thanks @adegeo

PR #10164, opened 2024-03-31
Reverted the changes in f4bf1f5

lextm:
1. Since wfastcgi is no longer actively maintained by Microsoft, HttpPlatformHandler should remain the recommended approach in this article.
2. It is more common to call this IIS out-of-band module "HttpPlatformHandler", instead of "HttpPlatform". You can find that formal name from IIS documentation like https://www.iis.net/downloads/microsoft/httpplatformhandler.
3. ASP.NET Core module is irrelevant. People often were confused due to an ambiguous announcement made by ASP.NET Core team. I wrote about that in https://docs.lextudio.com/blog/the-rough-history-of-iis-httpplatformhandler/#the-mist-of-aspnet-core.
4. HttpPlatformHandler alone can handle Python web apps very well, so there is no need to use an ASP.NET Core specific module (which is also bigger to load into memory due to ASP.NET Core in-process mode).

Thus, in this pull request I reverted all changes made in the earlier commit, revised HttpPlatformHandler name in references.

### DIFF
--- a/docs/python/configure-web-apps-for-iis-windows.md
+++ b/docs/python/configure-web-apps-for-iis-windows.md
@@ -25,38 +25,11 @@ When you use Internet Information Services (IIS) as a web server on a Windows co
 
 ## Set web.config to point to the Python interpreter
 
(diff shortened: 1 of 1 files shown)


ShannonLeavitt:
@cwebster-99

IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

cwebster-99:
#sign-off Had a discussion offline about the implications of this change and strictly recommending HttpPlatformHandler for the Python audience.

cwebster-99 (approved):

PR #10228, opened 2024-05-13
Fix the markdown rendering issue in the "Report a problem with the Visual Studio product or installer" documentation.

kartheekp-ms:
Fix the markdown rendering issue in the ["Report a problem with the Visual Studio product or installer"](https://learn.microsoft.com/visualstudio/ide/how-to-report-a-problem-with-visual-studio?view=vs-2022) documentation. As you can see in the screenshot below, every line below the `Record your actions to reproduce the issue` section is clickable. Clicking on the link does nothing. I think this rendering issue is confusing for customers. Additionally, internal product teams can't add a direct link to any section below the `Record your actions to reproduce the issue` topic while responding to a Visual Studio feedback ticket.

image

### DIFF
--- a/docs/ide/how-to-report-a-problem-with-visual-studio.md
+++ b/docs/ide/how-to-report-a-problem-with-visual-studio.md
@@ -45,7 +45,7 @@ Here are the steps to report a problem.
     ![Take a screenshot](../ide/media/feedback-screenshot.png)
     *Only Microsoft engineers can see the screenshot*
 
-1. <a name="trace"/>Record your actions to reproduce the issue.
+1. Record your actions to reproduce the issue.
 
    One of the best ways to help the Visual Studio engineering team solve the problem is to provide a trace and heap dump files for them to look through. You can do that by recording the steps that resulted in the bug. A screenshot will be captured every time the mouse is clicked, but keyboard entry will not record screenshots.
    1. Click **Start recording**. Wait a moment for the permissions prompt.


kartheekp-ms:
#sign-off

kartheekp-ms:
> Invalid command: '#sign-off'. Only the assigned author of one or more file in this PR can sign off. @jasonchlus

Please consider updating the PR template https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/pull_request_template.md?plain=1 may be by removing `When your PR is ready for review, add a comment with the text #sign-off to the Conversation tab.` sentence.

Jak-MS:
@jasonchlus 
- Can you review this PR?


#label:"aq-pr-triaged"

Jak-MS:
actually the change is small. approved. thanks @kartheekp-ms

jasonchlus:
@Jak-MS Sorry I didn't comment but I did review this, it is a very small change all good from my end :)

zivkan:
@kartheekp-ms @Jak-MS @jasonchlus I used to provide customers this link, to take them straight to the point about using the step recorder, and now the deep link doesn't work: https://learn.microsoft.com/en-us/visualstudio/ide/how-to-report-a-problem-with-visual-studio?view=vs-2022#trace

It's hard enough to get customers to actually use this "record your steps" functionality, very few customers reporting issues to my team actually use it, so I felt like it was useful being able to point them directly to the step in the process that I'd like them to use.

jasonchlus:
@zivkan would this link work: https://learn.microsoft.com/en-us/visualstudio/ide/how-to-report-a-problem-with-visual-studio?view=vs-2022#when-further-information-is-needed

zivkan:
@jasonchlus it doesn't emphasise the step recorder, which is my intent when I provide the URL to customers. The customers would have already seen the "attach system logs and recordings" from when they originally submit the feedback, but they either missed or skipped the step recorder then, so I'm trying to make it painfully obvious to them what exactly I'd like them to do.

While this is also a problem with the `#trace` link, since there are so many images in the docs page, and the rendered HTML doesn't put image sizes on any of the image tags, when I click the link you provided to me, the browser scrolls down to where the anchor is when the page doesn't have any images loaded, but then quickly/immediately loads images, so the browser no longer shows me the heading anywhere on my screen. Only by knowing enough about HTML and the browser URL bar, I can see it's looking for something, and I can scroll around.  Or close the browser tab and open the link a second time, since the browser now has the images cached. I don't know how to contact the docs team, but if you can pass on feedback that I think the pipeline that converts markdown into html should include height metadata on all img tags, so that deep links don't take you to the wrong place when deep links are followed, that'd be great.

I've given this feedback in the past, from which the outcome was adding that `#trace` anchor, but I'd personally prefer if there was a whole docs page dedicated to nothing other than the steps recorder, starting with a quickstart, and then deep dives into any features it has (it used to allow customers to add notes, not any more, plus the Windows Steps Recorder has a message saying it's going to be removed in the future. I'm not sure what the VS Feedback team are planning when this happens).

Having said this, I've had close to zero luck getting customers to actually provide recordings. At least 99% of the time, they never reply and their feedback gets auto closed as not enough information. Most of the time when they do reply, they still don't provide a recording and try to convince me again that they're experiencing this, but without providing me with anything to reproduce the issue myself. So maybe my feedback won't result in any measurable outcome anyway. 🤷

Jak-MS:
@zivkan @kartheekp-ms @jasonchlus - there's no reason we can't add the anchor back into the document. many articles have them.  The issue with this was (I believe) that the tag wasn't closed. I've created a PR to address this: https://github.com/MicrosoftDocs/visualstudio-docs/pull/10265

PR #10237, opened 2024-05-16
Update to gender neutral language

Jonny-vb:
Replace uses of "he" and "him" with gender neutral terms


Court72:
Thanks, @Jonny-vb!

PR #10239, opened 2024-05-16
Specify exact project in -targets query

rainersigwald:
If a `.sln` file is in the current working directory, `-ts` will return an error. Instead guide users to specifying a specific project to query about its targets.


rainersigwald:
@ghogen Is there an established pattern to the "insert your specific project here" in sample command lines that I should use instead?

Jak-MS:
@ghogen 
- Can you review this PR?


#label:"aq-pr-triaged"

ghogen (approved):

ghogen:
@rainersigwald. The formulation you're using is suited for the purpose here, so I think we should accept it and publish. We also can use substitution elements with delimiters. We used to use angle brackets, but those don't always work well since they can get interpreted as HTML or XML tags. So I now prefer to use curly braces, as in `{path to project}`.

PR #10306, opened 2024-07-10
Fix grammar in usingtask-element-msbuild.md

MiYanni:
## Summary

I was looking over this document and I reread this sentence several times, trying to understand what the intent was.

> Because there is no `Runtime` or `TaskHost` specified, the task will be executed the MSBuild process, in the runtime and architecture that happen to be running for a given build.

I believe it meant that "the task will be executed in the current MSBuild process", but I just added `in` since I wasn't exactly sure. Feel free to modify the PR if there is a better phrasing for clarity.


Jak-MS:
@ghogen 

#label:"aq-pr-triaged"

ghogen (approved):
@MiYanni  Thanks! I believe you're reading it and correcting it in the right way.

PR #10347, opened 2024-08-14
Use ToDouble() instead of ToInt32() for double type variable

cmkaya:
The variable `num2` is of type double, but the code was using `ToInt32()` to convert it. This can lead to potential loss of precision or rounding errors. The method should be updated to use `ToDouble()` to properly handle the double type.


cmkaya:
I noticed that the screenshot on line 407 needs to be updated. Currently, it displays the `ToInt32()` conversion, but it should reflect the use of `ToDouble()` instead, as the variable `num2` is of type `double`.

Court72:
@anandmeg

#label:"aq-pr-triaged"

anandmeg:
@cmkaya, thanks for the updates to help improve our docs!
#sign-off

PR #10350, opened 2024-08-15
fix: update Program.cs to use top-level statements for vs-2022

cmkaya:
This commit updates the structure of `Program.cs` for _vs-2022_ to use top-level statements, which is consistent with .NET 8.

Previously, the `Program` class was defined using the traditional way, which is not compatible with top-level statements.

### DIFF
--- a/docs/get-started/csharp/tutorial-console.md
+++ b/docs/get-started/csharp/tutorial-console.md
@@ -606,89 +606,85 @@ Let's get started.
 
     ```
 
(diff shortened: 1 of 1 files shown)


ShannonLeavitt:
@anandmeg

#label:"aq-pr-triaged"

anandmeg:
The tutorial is updated for top-level statements in VS 2022. When you create a new C# console project in VS 2022, the default code is `Console.WriteLine("Hello, World!");`
image

PR #10358, opened 2024-08-19
Update copilot-chat-context.md

amytnguyen01:
Made very small changes to ensure consistent naming of Copilot as "experience" instead of extension ("extension" is to be only used when referring to 17.9 and below specific instructions for installation)


anandmeg (approved):

anandmeg:
#sign-off

PR #10427, opened 2024-10-01
Update how-to-report-a-problem-with-visual-studio.md

jasonchlus:
Updated numbers


jasonchlus:
@ghogen can you approve and merge this basic change :)

ghogen:
@jasonchlus That anchor link is messing up the automatic numbering. There is a better fix. Just putting the numbers in manually will still produce a broken list (note the indentation in the live article is anomalous at that step). I'll create a new PR in the private repo to deal with this.

PR #10459, opened 2024-10-28
Update UdtKind C enumeration code

Michael-K-Stein:
A new UdtKind has been added to the UdtKind enumeration - "UdtTaggedUnion". This change is reflected in the table below but does not appear in the C example code under "Syntax"


Mikejo5000 (approved):
Thanks for the update!

Mikejo5000:
#sign-off

PR #10478, opened 2024-11-11
Fix project name in adding-a-tool-window.md

saranshsaini:
I think this is a typo, the first step says to call the project "FirstToolWin" but then throughout the project is referenced as FirstToolWindow. For example the second step under the "Add a toolbar to the tool window" section says find the node with name "guidFirstToolWindowPackageCmdSet" but that should be "guidFirstToolWindPackageCmdSet" if the project were called FirstToolWin, right?

Caused confusion for me as a first time VS extension developer so just wanted to clarify and fix this. Lmk if I am missing something.


Jak-MS:
@maiak 

#label:"aq-pr-triaged"

PR #10484, opened 2024-11-13
Fix incorrect enum value name in extended documentation.

Michael-K-Stein:
Seems that the line was copy-pasted from the one above and the word was missed


grantri (approved):

grantri:
#sign-off

PR #10487, opened 2024-11-14
Update and rename msbuild.md to msbuild.exe

jimmyjames337:


Jak-MS:
@ghogen fyi

Jak-MS:
@jimmyjames337 - not sure what you're trying to do with this PR, but we can't change the file name or file extension. thanks

PR #10493, opened 2024-11-15
Update example usage for IDiaSymbol::get_value

Michael-K-Stein:
I believe that using `VariantInit` from oleauto.h is more future-proof rather than manually initializing the VARIANT's fields in case the internal implementation of VARIANT/VARIANTARG changes.

Regarding the "see also" added at the end: I see that it refers to "previous versions" topic and there is a warning banner when entering the page. Not sure if it is better this way or if we should refer readers to [VariantInit function (oleauto.h)](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-variantinit) directly. On the other hand, that page refers readers to [Variant Manipulation Functions](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/automat/variant-manipulation-functions)

### DIFF
--- a/docs/debugger/debug-interface-access/idiasymbol-get-value.md
+++ b/docs/debugger/debug-interface-access/idiasymbol-get-value.md
@@ -44,7 +44,7 @@ The supplied VARIANT must be initialized before it is passed to this method. For
 void ProcessValue(IDiaSymbol *pSymbol)
 {
     VARIANT value;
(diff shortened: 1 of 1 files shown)


Court72:
@Mikejo5000

#label:"aq-pr-triaged"

Mikejo5000 (approved):
Thanks for the contribution!

Mikejo5000:
#sign-off

PR #10495, opened 2024-11-15
Add CV_MOD_UNALIGNED to examples

Michael-K-Stein:
Double checked this with a quick compilation.

Given this struct compiled with `/Od`:
```C

typedef struct FieldsWithModifiers_s
{
    volatile long volatileField;
    const long constField;
    char padding1;
    _UNALIGNED long unalignedField;
    char padding2;
} FieldsWithModifiers;
```

We get that the _type_ of "unalignedField" is unaligned. That is:
"unalignedField" is a `SymTagData` symbol, its `get_type` returns a `SymTagBaseType`, and when performing `get_unalignedType` we get `true`. Calling `get_unalignedType` on "unalignedField" of tag `SymTagData` actually returns `false`. Not sure how intuitive it is, but should probably be mentioned in the docs.

### DIFF
--- a/docs/debugger/debug-interface-access/cv-modifier-e.md
+++ b/docs/debugger/debug-interface-access/cv-modifier-e.md
@@ -58,6 +58,7 @@ typedef enum CV_modifier_e
 | `CV_MOD_INVALID` | Invalid modifier (unused) |
 | `CV_MOD_CONST` | C++ `const` |
 | `CV_MOD_VOLATILE` | C++ `volatile` |
+| `CV_MOD_UNALIGNED` | C++ `__unaligned` |
 | `CV_MOD_HLSL_UNIFORM` | HLSL uniform |
 | `CV_MOD_HLSL_LINE` | HLSL line |
 | `CV_MOD_HLSL_TRIANGLE` | HLSL triangle |


Court72:
@grantri

#label:"aq-pr-triaged"

grantri (approved):
#sign-off

grantri:
#sign-off

grantri:
@Michael-K-Stein - Maybe this is obvious to me only, but in C++ unaligned is a property of the type system, not of fields or data. so that is why `get_unalignedType` will always return false on `SymTagData`
Also the enum is exclusive to PDB's produced by fxc.exe (as mentioned in the topic's note) and thus you will not see this value in PDBs produced by MSVC. I was in a rush to document all of these APIs that have existed for quite a while, but were previously undocumented. `IDiaSymbol::get_modifierValues` is not just an alternate representation of `get_unalignedType`, but actually represents different fields within the PDB. Thus since fxc.exe is deprecated, all of these values and APIs are effectively dead code now.

Michael-K-Stein:
@grantri - Makes sense. Probably worth mentioning or completely removing `get_unalignedType`, `get_constType`, `get_volatileType` from the "Properties" section in [Data (Debug Interface Access SDK)](https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/debugger/debug-interface-access/data-debug-interface-access-sdk.md). Not sure about the use-case/meaning of these methods otherwise...?
Tagging @Mikejo5000 as well on this

grantri:
@Michael-K-Stein 
1. We can't remove the documentation because the APIs still exist and are functional.
2. I think you misunderstood me: `get_unalignedType`, `get_constType`, and `get_volatileType` are the produced by C++ and much more common amongst other compilers too.  It is `get_modifierValues` and `CV_MOD_UNALIGNED`, `CV_MOD_CONST` and `CV_MOD_VOLATILE` that are only produced by fxc.exe
3. You are correct that the docs could be clearer (e.g. " `TRUE` if the data is marked as volatile.  " could be better worded as " `TRUE` if the `SymTagType` is marked as volatile"). I'll see if I can work with @Mikejo5000 to get that cleaned up.
Rather than spamming this thread you can email me directly: same username at microsoft.com
Also please continue to post issues to the docs and offer suggestions or improvements.

PR #10501, opened 2024-11-19
Update attach-to-process-running-in-docker-container.md

sbonds:
I tried to keep it concise, but the idea is that this process needs to be done first:

https://learn.microsoft.com/en-us/cpp/linux/connect-to-your-remote-linux-computer?view=msvc-170

### DIFF
--- a/docs/debugger/attach-to-process-running-in-docker-container.md
+++ b/docs/debugger/attach-to-process-running-in-docker-container.md
@@ -69,7 +69,7 @@ To attach to a running process in a Linux Docker container:
 
       - To connect to a remote machine via SSH:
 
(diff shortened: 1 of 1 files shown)


Jak-MS:
@Mikejo5000 

#label:"aq-pr-triaged"

Mikejo5000:
@sbonds Thanks for the contribution! It looks like you can access the **Connect to Remote System** dialog either from **Select Docker Container** > **Add** (as currently described in the article), or from **Tools > Options > Cross-Platform > Connection Manager**. Is there a reason you think the Connection Manager step is requird here? fyi @gregg-miskelly

PR #10531, opened 2024-12-11
Remove VS 2017 from description metadata

ghogen:


Jak-MS:
trusting edit by @ghogen 

fyi @Mikejo5000

PR #10568, opened 2025-01-12
Update code sample on configure-unit-tests-by-using-a-dot-runsettings-file.md

abatishchev:
Updated the C# code sample in section TestRunParameters: it's either a field or a variable, not both at the same time.


Mikejo5000 (approved):

Mikejo5000:
#sign-off @abatishchev Thanks for the correction!

PR #10587, opened 2025-01-24
Fix typo

changeworld:
IPSec -> IPsec


ttorble:
Looks good, thanks @changeworld

PR #10590, opened 2025-01-24
Fix typo

changeworld:
vue CLI -> Vue CLI


ttorble:
Looks good, that's @changeworld

PR #10750, opened 2025-05-13
response files: correct component element, remove github as no longer a valid id

dpvreony:
Greetings

* This changes the response.json samples to have "components" instead of "add"
* it also removes the github component as it's no longer valid

Thanks


v-regandowner:
@anandmeg - Can you review the proposed changes? 
	
#label:"aq-pr-triaged"

anandmeg:
Converting the example response.json files to use *components* is incorrect. Response.json files use *add*.

dpvreony:
@anandmeg the documentation is wrong, the bootstrap installer throws an exception that the previous response.json using `add` is invalid

PR #10782, opened 2025-06-02
Fix typo: retreive -> retrieve

changeworld:


v-regandowner:
Looks good; thanks @changeworld

PR #10816, opened 2025-06-12
DONT ACCEPT THIS PR: referenced section "Get Involved" doesn't exist

AdmiralSnyder:
this is for adding a note that the docs point to something nonexistet
Sadly, Issues are not enabled on the repo (please do!)

### DIFF
--- a/docs/ide/whats-new-visual-studio-docs.md
+++ b/docs/ide/whats-new-visual-studio-docs.md
@@ -133,6 +133,7 @@ Welcome to what's new in Visual Studio docs. The following sections provide deta
 ### Community contributors
 
 The following people contributed to the Visual Studio docs during this period. Thank you! Learn how to contribute by following the links under "Get involved" in the [what's new landing page](index.yml).
+PLEASE DON'T INSERT THIS, it's a COMMENT: In lieu of another way of providing feedback: there is no "Get involved" section in the what's new landing page :-( - please enable Issues on the VS docs github repo
 
 - [changeworld](https://github.com/changeworld) - Takashi Takebayashi ![12 pull requests.](https://img.shields.io/badge/Merged%20Pull%20Requests-12-green)
 - [AdmiralSnyder](https://github.com/AdmiralSnyder) - Alexander Gayko ![1 pull requests.](https://img.shields.io/badge/Merged%20Pull%20Requests-1-green)


v-regandowner:
Closing this PR to prevent accidental merging. 

@ghogen - Please note the commented suggestion.

ghogen:
@AdmiralSnyder @v-regandowner 
Thanks for the feedback - I have a fix in progress for the issue you raised.

Note: even though GitHub issues are not available on the repo now, we still can get feedback if you use the Feedback controls on any page.

AdmiralSnyder:
@ghogen thanks for the reply :-)
> Note: even though GitHub issues are not available on the repo now, we still can get feedback if you use the Feedback controls on any page.
are you referring to this?
image
which points to https://developercommunity.visualstudio.com/VisualStudio/suggest, out of context, and far, far away from github... - not a good experience, if you ask me.

BUT the _actual reason why i wanted to file an issue was:
can we include the fact that the .net SDK version is bumped in VS releases, when that's the case?
this week, 17.14.5 came out and the SDK jumped from 9.0.300 to 9.0.301 - this means that updating prevents a solution from compiling, when the version is pinned in the global.json - and i didn't find a way to provide feedback on the release notes page, at all.

ghogen:
@AdmiralSnyder

Re: Feedback

I was referring to the text box that appears when you press either Yes or No, not the "Provide product feedback". But in any case, your feedback is noted. The Feedback mechanism was changed a few years ago. There are some blog posts discussing this change. See https://techcommunity.microsoft.com/blog/microsoftlearnblog/announcing-a-new-way-to-give-feedback-on-microsoft-learn/4027635.  Note that GitHub issues is still used for open-source projects like .NET itself.

Re: SDK versions

It sounds like you're referring to the .NET SDK and recommending that the release notes page, or some other official page, state explicitly when the .NET SDK version was updated. That feedback could go via the Product Feedback link that takes you to Developer Community, as it will reach the appropriate audience there. It does seem like a reasonable suggestion.

AdmiralSnyder:
@ghogen this
https://developercommunity.visualstudio.com/t/Include-NET-SDK-versions-in-release-not/1694079
is what i lovingly call devcom limbo - where ideas go to die.
(that suggestion from march 2022 is under review since a week after its creation, and is basically what i was intending to suggest. Please don't tell me that my additional upvote, 3 years after, would change anything there...) the docs are edited in github, so that is where issues regarding the docs should go, imo.

ghogen:
@AdmiralSnyder 
Indeed, we always regret that we can't implement every suggestion, and as you know there are many business considerations and tradeoffs in any backlog of possible work.

PR #10901, opened 2025-06-30
Fix typo: propert page -> property page

changeworld:


v-dirichards:
Looks good, thanks @changeworld!

PR #11031, opened 2025-09-02
Clarify meaning of 'letter' in valid property name characters description

baronfel:
Small change to resolve some ambiguity.


ghogen:
Thanks, @baronfel !

PR #11312, opened 2026-04-21
Replace double hyphens with em dashes

ihor-kitsa:


v-dirichards:
Looks good, thanks @ihor-kitsa!

PR #11315, opened 2026-04-22
Delete superfluous "to"

ihor-kitsa:


v-regandowner:
Verifiable typo fix, thanks!

PR #11445, opened 2026-09-10
Enhance template example clarity

frogcrush:
Clarified the example for templates by specifying the default name in the New dialog and the location of the C# class template.

If you don't have the `xmlns` in there, Visual Studio will show an error when you try to use a custom template.

Additionally, I feel like `ReplaceParameters="true"` being in there can help add clarity to whether or not parameters are replaced in the template file. This held me up for a bit when creating my own template. Having it as `true` by default in the example feels like the safer option.


v-regandowner:
@RoseHJM 

#label:"aq-pr-triaged"

RoseHJM (approved):

RoseHJM:
[Tylor Pater (@frogcrush)](https://github.com/frogcrush): Thanks for the updates!

